PR #1, opened 2020-09-17
Fix scrutiny binary location

RealOrangeOne:
> ERROR: for web  Cannot start service web: OCI runtime create failed: container_linux.go:349: starting container process caused "exec: \"/scrutiny\": permission denied": unknown

This happens because it is a directory, the binary itself is in a different place


AnalogJ:
Oh this is great, thanks. I kinda threw together the non-omnibus dockerfiles, hadn’t had time to test yet

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #75, opened 2020-10-03
Notifications system

### LINKED ISSUE #25: Webhooks & notification system
- [x] custom scripts
- [x] webhooks
- [x] 3rd party services (discord/slack/pushbullet/pushover/etc)
- [ ] filter/disable notifications per drive.

### LINKED ISSUE #55: [FEAT] Test notification
I just setup scrutiny and like it. I tried to setup a notification and hopefully have included the config into docker correctly. But since my HDDs all work well at the moment (and I don't want that to change), **is there a way to provoke a notification**? Maybe on the CLI? Just to test that it'll work when it really detects something is wrong. Thanks!

AnalogJ:
- Note: does not support disabling notifications. Tracked by another issue.

closes #55 
closes #25

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,6 +135,7 @@ Check the [example.scrutiny.yml](example.scrutiny.yaml) file for a fully comment
 ## Notifications
 
 Scrutiny supports sending SMART device failure notifications via the following services:
+- Custom Script (data provided via environmental variables)
 - Email
 - Webhooks
 - Discord
--- a/README.md
+++ b/README.md
@@ -160,6 +160,47 @@ You can test that your notifications are configured correctly by posting an empt
 curl -X POST http://localhost:8080/api/health/notify
 ```
 
+# Debug mode & Log Files
+Scrutiny provides various methods to change the log level to debug and generate log files.
+
+## Web Server/API
+
+You can use environmental variables to enable debug logging and/or log files for the web server:
+
+```
+DEBUG=true
+SCRUTINY_LOG_FILE=/tmp/web.log
+```
+
+You can configure the log level and log file in the config file:
+
+```
+log:
+  file: '/tmp/web.log'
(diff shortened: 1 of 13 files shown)


kbilkis:
hey! thanks for implementing this!

I do realise this is a work in progress, but fyi - to get discord notifications working I had to comment out all urls except for `discord` and `script` for whatever reason.

AnalogJ:
Hey @kbilkis That's expected actually. All the url's are just examples. As a user you'll need to uncomment out the services that you want to use for notifications. 
Is there a reason you expected it to work differently?

kbilkis:
In my use case I wanted to have only the `discord` notifications, so I commented out all others, but `discord` wasn't working. After some trial-and-error I've discovered that I specifically need to uncomment the `script` 'url', which doesn't make sense at all, as it didn't even have a valid directory set I believe. But without that line `discord` doesn't work :/

AnalogJ:
Oh, that definitely sounds like an error. Can you open a new issue for that @kbilkis ?
I want to track that bug separately from this closed PR.

kbilkis:
> Oh, that definitely sounds like an error. Can you open a new issue for that @kbilkis ?
> I want to track that bug separately from this closed PR.

sure [issue 116](https://github.com/AnalogJ/scrutiny/issues/116)

README.md
```diff
@@ -132,6 +132,76 @@ We support a global YAML configuration file that must be located at /scrutiny/co
...
+- Hangouts
+- IFTTT
+- Join
+- Mattermost
+- Pushbullet
+- Pushover
+- Slack
+- Teams
+- Telegram
+- Tulip
```
tianon, line 152: Looks like this was a typo and you probably meant Zulip here? 😅


```suggestion
- Zulip
```

PR #89, opened 2020-10-08
[WIP] working on a true static binary.

### LINKED ISSUE #82: [BUG] Pre-compiled app depends on a specific glibc version, so it basically only works on debian buster
Going through the [MANUAL_INSTALL.md](https://github.com/AnalogJ/scrutiny/blob/master/docs/INSTALL_MANUAL.md) file, it states:

> Since the webapp is packaged as a stand alone binary, there isn't really any software you need to install other than glibc which is included by most linux OS's already.

This isn't true. I tried to set up the application on ubuntu server 18.04, and it will not run, as the binary depends on glibc 2.28, and ubuntu server 18.04 ships glibc 2.27. Updating to 20.04 gives you glibc 2.30.

Digging around, it looks like the the distro that ships glibc 2.28 is debian buster, so I'm guessing this is the platform you built the software on.

Basically, it could probably be made to work, but would require recompiling the application.

The documentation should probably reflect the platform limitations.

Strangely, the collector doesn't seem to suffer from these issues. Not sure why.

AnalogJ:
fixes #82

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -27,7 +27,7 @@ jobs:
         env:
           GOOS: linux
           GOARCH: amd64
-          CGO_ENABLED=0
+          CGO_ENABLED: 0
         run: |
           apt-get update && apt-get install -y file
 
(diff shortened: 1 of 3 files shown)


AnalogJ:
fixes #82

francescayeye:
@AnalogJ I see that you need dynamic linking for both net package and sqlite dependency: for the latter `-tags sqlite_omit_load_extension` (I haven't checked but I don't think you are using extension loading in sqlite)

ref: https://golangnews.org/2020/04/statically-compiling-go-programs-2/

AnalogJ:
Hey @aspacca 
Thanks for the tip! Yeah, I think that worked. I also had to add some `--tags` but it looks like its building a true static binary now. I'm going to test it out and then roll my changes out to other arch and os flavors.

PR #103, opened 2020-10-16
Improve humanizeHours

ryck:
Replaces custom _humanizeHours_ function by [HumanizeDuration](https://evanhahn.github.io/HumanizeDuration.js/) to provide more accurate values.

PS: No one asked for this, and I am more comfortable with React, so don't feel bad and decline it if you think it doesn't add value, is not worthy or there are better ways to do it 😅

**Screenshots**
Dashboard
image

Detail
image


AnalogJ:
Nice, this is great! 🎉 
Ripping out my kludgy code is probably a good thing :D

PR #105, opened 2020-10-17
Use StringFlag for COLLECTOR_HOST_ID

ryck:
I have zero idea about Go, but testing the Host ID feature, I realised I couldn't use the  `COLLECTOR_HOST_ID` environment variable, and looks like it's because of this, but please double check!


AnalogJ:
haha, oooof. That's definitely a copy-paste error! Thanks for the PR! 🎉

PR #130, opened 2020-12-19
adding jobber socket for easier debugging

AnalogJ:


AnalogJ:
no longer relevant. using Cron.

PR #132, opened 2020-12-27
Change temperature graph to 24-hour formatting

### LINKED ISSUE #131: [BUG] Dashboard graph labels format the date strangely
**Describe the bug**
In the temperature graph at the bottom of the dashboard, the dates are formatted very strangely. Times go from hour 01 to hour 12, then wrap around (12:59:59 is followed by 01:00:00). The dates wrap around at one of the boundaries between 11:59:59 and 12:00:00. I suspect this all because it's using a 12 hour format specifier rather than a 24 hour one. The times themselves appear to be otherwise correct.

**Expected behavior**
Hours in the labels should be 24-hour formatted (i.e. go from 00 to 23) and days should change at the 23:59:59 -> 00:00:00 boundary. Or perhaps AM/PM should be displayed. I'd rather have 24-hour formatting, personally.

**Screenshots**

https://user-images.githubusercontent.com/175827/103173488-3beff880-4853-11eb-950d-c4d80342e75d.mov

(Yes, I know, I just bought some aftermarket fans because of this graph 😅)

telyn:
Fixes #131

### DIFF
--- a/webapp/frontend/src/app/modules/dashboard/dashboard.component.ts
+++ b/webapp/frontend/src/app/modules/dashboard/dashboard.component.ts
@@ -139,7 +139,7 @@ export class DashboardComponent implements OnInit, AfterViewInit, OnDestroy
             tooltip: {
                 theme: 'dark',
                 x    : {
-                    format: 'MMM dd, yyyy hh:mm:ss'
+                    format: 'MMM dd, yyyy HH:mm:ss'
                 },
                 y    : {
                     formatter: (value) => {


TheCatLady:
Would actually be nice to instead allow setting the date format string via an env var 😉  I personally prefer 12-hour notation w/ AM/PM.

telyn:
> Would actually be nice to instead allow setting the date format string via an env var 😉 I personally prefer 12-hour notation w/ AM/PM.

That’s fair! I’m not really up on my angular and not a regular contributor to this project so don’t have the confidence/knowledge to sort that out. I hope we can both agree that this PR is better than the behaviour before, where it was 12-hour without the AM/PM marker!

AnalogJ:
🥳  thanks for the PR @telyn! LGTM, merging. 

@TheCatLady Can you create a separate issue to track customizing the date format string?

PR #139, opened 2021-01-11
Add aarch64 Dockerfile

andreipoe:
Hello,

I have modified the Dockerfile to build on aarch64 systems. The only major change from amd64 is that we can't use a release build of jobber, so we build it from source in a separate container. For S6, I've only changed the URL to point to their aarch64 release.

This allows users that want to compile from source to build their own image locally:

```
docker build -t scrutiny -f docker/Dockerfile.aarch64 .
```

The original (amd64) Dockerfile is left untouched.

Would you be interested in integrating?

Thanks!


AnalogJ:
I'd be happy to integrate, though I'd like to keep a single Dockerfile, and just add support for a multi-architecture build using the Github Action & build-params. Is that something you'd be willing to contribute?

AnalogJ:
Hey @andreipoe, now that we've moved back to `cron` as our scheduler, do we still need this PR?
Does scrutiny build for you on an `aarch64` system?

PR #210, opened 2021-12-27
Added tzdata package

John-Appleseed:
Fixed missing timezone data by adding the tzdata package to apt-get.
With tzdata installed, the docker environment variable `TZ` will set the correct timezone during the init phase.


John-Appleseed:
Temporary fix is to share the host's zoneinfo to the container via `/usr/share/zoneinfo:/usr/share/zoneinfo:ro`
```
services:
  scrutiny:
    container_name: scrutiny
    volumes:
      - /usr/share/zoneinfo:/usr/share/zoneinfo:ro
```

PR #219, opened 2022-03-26
update go dependencies

taigrr:


AnalogJ:
Thanks for the PR @taigrr 
I'm actually working on a refactor based on InfluxDB which will change the dependencies (and update them) so this will be unnecessary. Thanks for the PR though! 🥳

PR #243, opened 2022-05-11
Add Ansible instructions

Zorlin:
This adds instructions on how to use the Ansible playbook for Scrutiny to automatically install it, including InfluxDB2.

(A future version of the Ansible playbook will also offer the option to choose between "manual" and "Docker" install styles for each host)


AnalogJ:
Awesome, thanks for the support & PR! 🥳 🎉

PR #272, opened 2022-05-30
Create INSTALL_PFSENSE.md

BadCo-NZ:
As requested by @AnalogJ, let me know if you spot any errors.


AnalogJ:
if you have any screenshots, that would be helpful as well, but we can make that a different PR. 

Thanks for the PR @BadCo-NZ! 🎉

BadCo-NZ:
Yeah I wanted to keep it all in the CLI to avoid screenshots, but there wasn't an easy way to do the cronjob.

I'll look at adding a screenshot soon :)

PR #303, opened 2022-06-15
Create INSTALL_COLLECTOR_SYNOLOGY_AARCH64.md

SiM22:
A little walk through to get the collector running on Synology


AnalogJ:
This is great, thanks @SiM22 !
I may rename the document and add some additional links, but this is exactly what we need. 

🎉

PR #350, opened 2022-08-03
Add missing setup command for udev troubleshooting

MattKobayashi:
I just learned about a command that's required after installing eudev on Alpine Linux to ensure udev updates are triggered on boot. I've added this to the TROUBLESHOOTING_UDEV doc.


AnalogJ:
Thanks for the update @MattKobayashi!

Issue #229, opened 2022-05-07
[FEAT] Helm Chart / k8s Deployment


somerandow:
**Is your feature request related to a problem? Please describe.**

Discovered this tool from a reddit post. It works fantastic on desktop, and as a user that leverages k8s for their homelab/server setups, it would be awesome to see this tool work in that environment as well.

**Describe the solution you'd like**

Reading through the documentation, it seems like this project already provides separate images for the collector and web frontend.
As far as I can tell, the only things needed to really support this type of deployment would be a Helm chart to ease/automate deployment, and the following issues:
#227 
#226 

I think #226 would be the most important here. K8s can handle spinning up the collector on every node, but not all nodes may be the same, so being able to wildcard or dynamically add devices would be a welcome feature.

**Additional context**

Don't have a ton of extra free time, but would be happy to help through a PR or testing

Zorlin:
+1, might have a crack at a Helm chart at some point.

AnalogJ:
I would definitely appreciate a PR to add a helm chart. 
However, I did some quick analysis of `device-cgroup-rule` in #226, and it doesnt correctly pass in existing devices, only new devices. 

the only solution i can think of is volume mounting `-v /dev:/dev` and that is really sub-optimal. I'm hoping one of you has a better idea!

vtmocanu:
does anyone have a working recent helm for scrutiny and can share it?

kaysond:
This issue is being closed automatically as a part of an effort to streamline and prioritize development effort. If it is still relevant, please reply to this comment and we will convert it to a discussion.

PR #363, opened 2022-08-28
helm: add helm chart

nadiamoe:
This PR adds a basic helm chart, which allows easy deployment of scrutiny in Kubernetes.

This is a pretty rough first iteration, but is as far as I can tell funcitonal and allows certain degree of customization.

There are some rough edges I've left as review comments, just to doublecheck what I'm doing makes sense.

Please let me know what you think!


charts/scrutiny/charts/collector/templates/daemonset.yaml
```diff
@@ -0,0 +1,61 @@
...
+            - secretRef:
+                name: {{ include "collector.fullname" . }}-credentials
+          env:
+            {{- range $key, $value := .Values.env }}
+            - name: {{ $key }}
+              value: {{ $value | quote }}
+            {{- end }}
+          resources:
+            {{- toYaml .Values.resources | nindent 12 }}
+          volumeMounts:
```
nadiamoe, line 40: From what I can tell, the collector writes stuff in `/opt/scrutiny`.
Good K8s hygiene mandates the root of the container to be read-only, and all locations where temporary files are written to be specific EmptyDir mount points. Is there any other location where this docker image writes to?

AnalogJ:
woah, this is great, thanks! 🥳 
I should be able to give you a detailed review this weekend.

charts/scrutiny/charts/web/templates/hpa.yaml
```diff
@@ -0,0 +1,28 @@
+{{- if .Values.autoscaling.enabled }}
```
AnalogJ: We should probably remove this functionality completely. 
Scrutiny uses a sqlite database under the hood (in addition to the InfluxDB). The sqlite file is used to store device metadata. Sqlite isn't designed for multiple writers, so scaling the web container horizontally wouldn't be safe.

charts/scrutiny/charts/collector/templates/daemonset.yaml
```diff
@@ -0,0 +1,61 @@
+apiVersion: apps/v1
```
AnalogJ, line 1: I'd love to use Kubernetes's built in cron-scheduling, instead of my kludgy cron-in-docker solution if we can. 
I've used https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/ in the past, but I'm not sure if it has the capabilities to force a container on every node. Is that something you're familiar with?

charts/scrutiny/Chart.yaml
```diff
@@ -0,0 +1,29 @@
...
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.1.0
+
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
+# It is recommended to use it with quotes.
+appVersion: "1.16.0"
```
AnalogJ: this should probably match the scrutiny version right? v0.4.x.
We'll probably want to tweak our CI/CD so that new releases update this file automatically. I can handle that in a separate PR.

charts/scrutiny/charts/collector/Chart.yaml
```diff
@@ -0,0 +1,24 @@
...
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.1.0
+
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
+# It is recommended to use it with quotes.
+appVersion: "1.16.0"
```
AnalogJ: ditto ^

charts/scrutiny/charts/collector/templates/daemonset.yaml
```diff
@@ -0,0 +1,61 @@
...
+            - name: {{ $key }}
+              value: {{ $value | quote }}
+            {{- end }}
+          resources:
+            {{- toYaml .Values.resources | nindent 12 }}
+          volumeMounts:
+            - mountPath: /run/udev
+              name: udev
+      volumes:
+        - name: udev
```
AnalogJ, line 44: on K8s, the collector will probably need to run as `privileged` or mount the `/dev/` directory (which is basically the same thing). 

https://github.com/AnalogJ/scrutiny/issues/229

basically we need to make sure that the collector container has access to all the storage devices. When running docker manually, my README instructions state that users should pass each device individually using `--device` but that won't be possible in K8s. 


Thankfully the `web` container does not need this privileged access.

charts/scrutiny/charts/collector/values.yaml
```diff
@@ -0,0 +1,61 @@
...
+# Declare variables to be passed into your templates.
+
+env:
+  COLLECTOR_API_ENDPOINT: http://scrutiny-web:8080
+
+image:
+  repository: ghcr.io/analogj/scrutiny
+  pullPolicy: IfNotPresent
+  # Overrides the image tag whose default is the chart appVersion.
+  tag: master-collector
```
AnalogJ: we should probably lock this tag to a specific version (that matches the appVersion in the Chart.yaml)

[v0.5.0-web](https://github.com/AnalogJ/scrutiny/pkgs/container/scrutiny/32343361?tag=v0.5.0-web)

charts/scrutiny/charts/web/templates/hpa.yaml
nadiamoe: I originally thought that multiple instances of the `web` service were possible (each with its own db), but what you're saying makes a lot of sense. I'll get rid of this.

charts/scrutiny/charts/collector/templates/daemonset.yaml
nadiamoe: I completely agree, I do not particularly like the cron-in-docker thing. Unfortunately as far as I know, there is no reliable way to run a CronJob on all nodes :(
I've seen hacks here and there, like forcing a number of replicas == number of nodes and setting antiaffinity rules, but I don't think that's very reliable.

An alternative to the cron in docker could be a simpler image that just does:
```sh
#!/bin/sh
while [ true ]; do
    /opt/scrutiny/bin/something something
    sleep 6h
done
```

This script is simple enough to inline in the K8s manifest. WDYT?

charts/scrutiny/Chart.yaml
nadiamoe: Yep, I honestly left the defaults as a WIP.

I've had problems in the past with keeping this up to date on CI/CD, so if you can we can also strip it altogether. It is not required for it to be present.

charts/scrutiny/charts/collector/templates/daemonset.yaml
nadiamoe: Right now the collector runs with `privileged: true` and the web does not, so that should be ok. I am not sure about whether the `web` can run as a non-root user though.

`privileged: true` already mounts `/dev`, does `/run/udev` need to be mounted as well or can I get rid of it?

charts/scrutiny/charts/collector/values.yaml
nadiamoe: LGTM, if you agree with keeping the `appVersion` we can make the template default to `{{ .Chart.appVersion }}-web`.

nadiamoe:
Thanks for the in-depth review @AnalogJ!
On my side:
- [x] Get rid of the autoscaling functionality
- [x] Rewrite default tags to use `appVersion`

What I'd like to confirm:
- [x] Wether we need to mount `/run/udev`
- [x] If you're ok with maintaining `appVersion` in sync
- [ ] If you like the idea of replacing the cron-in-docker with the custom script inline in the manifest (no changes required to the docker image)

AnalogJ:
Hey @roobre 

- **Wether we need to mount /run/udev**
	- Yes, I think this is still required with `--privileged` 
- **If you're ok with maintaining appVersion in sync**
	- Yes, I'll tweak my packagr tooling to support multiple files - https://github.com/PackagrIO/docs
- **If you like the idea of replacing the cron-in-docker with the custom script inline in the manifest (no changes required to the docker image)**
	- Unfortunately I don't think we can use a simple custom script like you mentioned, some of our users definitely depend on the customization of the cron schedule. 

I hope that answers all your questions!

charts/scrutiny/charts/collector/templates/daemonset.yaml
nadiamoe: From your comment [here](https://github.com/AnalogJ/scrutiny/pull/363#issuecomment-1236157305) I understand there's nothing to change then :)

nadiamoe:
@AnalogJ 
We could expose an entry in `values.yaml` and use it to feed the script. For example:

```yaml
collector:
  period: 6h
```

And then:

```sh
while [ true ]; do
    /opt/scrutiny/bin/something something
    sleep {{ .Values.period }}
done
```

It is definitely not the same as cron, but might fill the gap and be more idiomatic than the cron.
Otherwise, we can just keep the cron in place.

charts/scrutiny/charts/collector/templates/daemonset.yaml
AnalogJ: yeah, sorry I must have skimmed over the `privileged` config

AnalogJ:
Hey @roobre I promise I'm not ignoring your hard work, I just got a bit distracted with some other projects -- https://www.reddit.com/r/selfhosted/comments/xj9rx7/introducing_fasten_a_selfhosted_personal/

I need to make some tweaks to my "bumpr" tool (to keep the versions in the helm manifest in sync with the scrutiny version) but this is looking pretty good.

Starttoaster:
Just wanted to say thanks @AnalogJ for the update, and @roobre for the hard work. Wanted to chime in as an interested 3rd party waiting for this to merge before I run scrutiny, personally.

issmirnov:
Hey folks,

Sorry to necromance old PR's, but I too am interested in deploying scrutiny. Is this helm chart in a good enough state to be used?

P.S. Fasten looks amazing, looking forward to running it on my k8s cluster too!

nadiamoe:
Hey @issmirnov,

I'm using this chart for my NAS and while some things UX-wise can definitely be improved it should work well for the most part.

charts/scrutiny/charts/web/templates/deployment.yaml
```diff
@@ -0,0 +1,77 @@
...
+            - secretRef:
+                name: {{ include "web.fullname" . }}-credentials
+          env:
+            {{- range $key, $value := .Values.env }}
+            - name: {{ $key }}
+              value: {{ $value | quote }}
+            {{- end }}
+          livenessProbe:
+            httpGet:
+              path: /
```
uhthomas:
```suggestion
              path: /health
```

https://github.com/AnalogJ/scrutiny/commit/f2856e0f2607237d9886fee8625070448d6de0c4

charts/scrutiny/charts/web/templates/deployment.yaml
```diff
@@ -0,0 +1,77 @@
...
+              port: http
+          readinessProbe:
+            httpGet:
+              path: /
```
uhthomas:
```suggestion
              path: /health
```

https://github.com/AnalogJ/scrutiny/commit/f2856e0f2607237d9886fee8625070448d6de0c4

charts/scrutiny/charts/web/templates/deployment.yaml
uhthomas, line 1: I am not sure this _can_ be a deployment. My understanding is that there is a sqlite database somewhere in `/opt/scrutiny/config` which is required. Issues occur if it's lost.

This should probably be a stateful set with one replica, or at least `/opt/scrutiny/config` should have persistence.

charts/scrutiny/charts/collector/templates/daemonset.yaml
uhthomas: I tried to make it use a read only file system, but it really freaks out. It want to write for entrypoints and cron job stuff. I think the collector would need significant changes to facilitate this.

charts/scrutiny/charts/collector/values.yaml
```diff
@@ -0,0 +1,62 @@
...
+podAnnotations: {}
+
+podSecurityContext: {}
+# fsGroup: 2000
+
+securityContext:
+  privileged: true
+  capabilities:
+    add:
+      - SYS_RAWIO
```
uhthomas, line 37: SYS_ADMIN is also required for nvme

charts/scrutiny/charts/web/templates/deployment.yaml
AnalogJ: correct, https://github.com/AnalogJ/scrutiny/blob/master/webapp/backend/pkg/config/config.go#L37 must be persisted

adepssimius:
This looks cool and would be awesome to have in my cluster. I've been using scrutiny for years on a homelab server and I hope I can continue to use it now that that I'm migrating to k8s.

nadiamoe:
I continue to use and be very happy with Scrutiny on k8s, but I'm trying to move away from Helm charts in general, so I won't be able to follow up on this PR.

There's good feedback here, if anyone wants to follow up and adapt my initial PR please feel free to!

PR #375, opened 2022-09-24
Support transient SMART failures

### LINKED ISSUE #374: [BUG] Hardware ECC Recovered incorrectly reported as disk failure
**Describe the bug**

This particular SMART attribute is expected to fluctuate up and down, especially during random IO, and is not indicative of disk failure. See [here](https://superuser.com/a/1511916/169872) for some background info. Also, it seems that for this attribute lower values are worse, not better.

**Expected behavior**
Scrutiny shouldn't report this as failure. Seagate's own SeaTools doesn't either.

**Screenshots**
See the last row.

Screenshot from 2022-09-24 17-13-58

dcelasun:
As discussed in #374 some SMART errors are transient and should not be treated as permanent.

This commit adds support for a configurable list of ATA SMART attribute IDs, failures of which will be treated as transient. Drive health history is still recorded and notifications are sent, but the device itself is not marked as failed.

Fixes #374.

--

@AnalogJ apologies for the cosmetic changes, they are made automatically by [goimports](https://pkg.go.dev/golang.org/x/tools/cmd/goimports). I can revert them if you'd like.

### DIFF
--- a/collector/pkg/config/interface.go
+++ b/collector/pkg/config/interface.go
@@ -20,6 +20,7 @@ type Interface interface {
 	GetInt(key string) int
 	GetString(key string) string
 	GetStringSlice(key string) []string
+	GetIntSlice(key string) []int
 	UnmarshalKey(key string, rawVal interface{}, decoderOpts ...viper.DecoderConfigOption) error
 
 	GetDeviceOverrides() []models.ScanOverride
--- a/collector/pkg/config/mock/mock_config.go
+++ b/collector/pkg/config/mock/mock_config.go
@@ -133,6 +133,20 @@ func (mr *MockInterfaceMockRecorder) GetInt(key interface{}) *gomock.Call {
 	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "GetInt", reflect.TypeOf((*MockInterface)(nil).GetInt), key)
 }
 
+// GetIntSlice mocks base method.
+func (m *MockInterface) GetIntSlice(key string) []int {
+	m.ctrl.T.Helper()
+	ret := m.ctrl.Call(m, "GetIntSlice", key)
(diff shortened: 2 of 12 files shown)


AnalogJ:
Thanks so much for the PR 🥳 

Apologies for taking so long to get to this, I've been a bit distracted by some other projects. 
TBH, I think I would have implemented this slightly differently. Rather than "ignoring" the failure at the attribute processing stage, I think i would allow it to mark the attribute entry as "failed" however, I would filter out the failure before setting the DeviceStatus -- which persists across multiple SMART collector runs. 

https://github.com/AnalogJ/scrutiny/blob/master/webapp/backend/pkg/web/handler/upload_device_metrics.go#L53-L59

Basically since the value fluctuates (and recovers) constantly, a failure at a single point in time should still be marked as a failure, but shouldn't cause the disk to the permanently marked as "bad"

Hopefully that makes sense?

dcelasun:
> Hopefully that makes sense?

It does. Let me make some changes :)

dcelasun:
Apologies for the delay on this. Now that I look at it again:

> Basically since the value fluctuates (and recovers) constantly, a failure at a single point in time should still be marked as a failure, but shouldn't cause the disk to the permanently marked as "bad"

Isn't this already the case? The `Status` field in the Smart` type refers to the device's status, not the attributes, and my PR simply doesn't update that field for transient failures.

enoch85:
Would be great if you could have another look at this. :pray:

rursache:
will this be ever merged?

kaysond:
@dcelasun - I think it might make sense to roll this into https://github.com/AnalogJ/scrutiny/pull/835

dcelasun:
> @dcelasun - I think it might make sense to roll this into #835

Sounds good to me

PR #390, opened 2022-11-03
feat: custom PUID/PGID

### LINKED ISSUE #362: [BUG] Permissions of Influxdb folder do not follow PUID/PGID set in the docker-compose file
I ran scrutiny with such docker-compose scheme on my raspberry pi 4b (4GB): 

```
    scrutiny:
        cap_add:
            - SYS_RAWIO
        ports:
            - 8123:8080
        environment:
            - PUID=${PUID}
            - PGID=${PGID}
            - TZ=${TIMEZONE}
        volumes:
            - /run/udev:/run/udev:ro
            - ${CONFIG_FOLDER}/scrutiny/config:/opt/scrutiny/config
            - ${CONFIG_FOLDER}/scrutiny/influxdb:/opt/scrutiny/influxdb
        devices:
            - /dev/sdb
        container_name: scrutiny
        restart: unless-stopped
        image: ghcr.io/analogj/scrutiny:master-omnibus
```

Since it was introduced the _influxdb_ database I now encounter the issue that the `/opt/scrutiny/influxdb` folder is not locally saved after my `PUID` and `PGID` settings (set in an `.env` file next to my docker-compose file). That is not the case for the `/opt/scrutiny/config` folder which instead respects the `PUID` and `PGID` settings. 

I tried adding `user: ${PUID}:${PGID}` in the docker-compose but in that way scrutiny does not event start. 

Do you know how to solve such issue?

adripo:
fix: #362

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -38,7 +38,9 @@ COPY /rootfs/etc/cron.d/scrutiny /etc/cron.d/scrutiny
 COPY --from=backendbuild /go/src/github.com/analogj/scrutiny/scrutiny /opt/scrutiny/bin/
 COPY --from=backendbuild /go/src/github.com/analogj/scrutiny/scrutiny-collector-metrics /opt/scrutiny/bin/
 COPY dist /opt/scrutiny/web
-RUN chmod +x /opt/scrutiny/bin/scrutiny && \
+RUN useradd -u 0 -U -d /config -s /bin/false abc && \
+    usermod -G users abc && \
+    chmod +x /opt/scrutiny/bin/scrutiny && \
     chmod +x /opt/scrutiny/bin/scrutiny-collector-metrics && \
     chmod 0644 /etc/cron.d/scrutiny && \
     rm -f /etc/cron.daily/* && \
--- a/docker/Dockerfile.collector
+++ b/docker/Dockerfile.collector
@@ -22,7 +22,9 @@ RUN apt-get update && apt-get install -y cron smartmontools ca-certificates tzda
 COPY /docker/entrypoint-collector.sh /entrypoint-collector.sh
 COPY /rootfs/etc/cron.d/scrutiny /etc/cron.d/scrutiny
 COPY --from=backendbuild /go/src/github.com/analogj/scrutiny/scrutiny-collector-metrics /opt/scrutiny/bin/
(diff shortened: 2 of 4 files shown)


AnalogJ:
hey @adripo 
Thanks for the PR! 🥳 

However I'm not quite sure this will work as intended.

- we should name our user `scrutiny` rather than `abc`
- `/config` doesnt exist in Scrutiny. our home dir should be `/home/scrutiny`
- the main problem is: the scrutiny-web, scrutiny-collector and influxdb processes will still run as root, not the new `abc/scrutiny` user. 

Hope that all makes sense, could you make those changes? Thanks again!

adripo:
Hi @AnalogJ, I just added the script in a hurry, but I am now working on different issues in this project so I will check all the point that you raised. I agree with your suggestions, but I have one doubt. If I run the main process with the new scrutiny user you will never be able to run it as root anymore because you cannot set id 0 to a different user, right?

AnalogJ:
We wouldn't start the main process (s6-overlay) with `root`, instead we would just start the child process as `scrutiny`. 

Once other thing I forgot to mention (that complicates this even further) is that `smartctl` needs to run as `root` from what I remember, even within the container.

So only the webapp and influxdb processes can be changed to run under the `scrutiny` user

AnalogJ:
Here's an old example for how to update the service files to run as a (non-root) user - https://github.com/just-containers/s6-overlay/issues/207

EkilDeew:
> Once other thing I forgot to mention (that complicates this even further) is that smartctl needs to run as root from what I remember, even within the container.

You could just add the `sudo` package and allow the `scrutiny` user to only use smartctl as sudo. ie (in the sudoers file or another file under /etc/sudoers.d) :

```
scrutiny ALL = (ALL) NOPASSWD:/bin/smartctl
```

kaysond:
@adripo - are you interested in working on this? there are some new ways to run things with s6 version 3

adripo:
Hi @kaysond! I will have a look at s6 version 3 because I am not up to date with the latest developments

kaysond:
> Hi @kaysond! I will have a look at s6 version 3 because I am not up to date with the latest developments

Thanks! Ideally, I'd like to support fully rootless containers, instead of with PUID/PGID, since there's nothing the container does that should strictly need root. S6 supports this (possibly with some env vars). It would also be great if we could support read only rootfs, which s6 can do with an env var

Issue #318, opened 2022-06-29
Context timing out on migrations (omnibus)


invisiblethreat:
**Describe the bug**
When attempting to migrate to an InfluxDB version of the migration fails with `Client.Timeout exceeded while awaiting headers`, which causes a `panic`. It's not entirely clear to me if this is a processing timeout, or a failure to connect to Influx that looks like a timeout rather than an auth failure. The issue resolved once I added the token from Influx into the config file.

**Expected behavior**
Migration completion.

**Log Files**
```
2022/06/29 11:02:54 Loading configuration file: /opt/scrutiny/config/scrutiny.yaml

 ___   ___  ____  __  __  ____  ____  _  _  _  _
/ __) / __)(  _ \(  )(  )(_  _)(_  _)( \( )( \/ )
\__ \( (__  )   / )(__)(   )(   _)(_  )  (  \  /
(___/ \___)(_)\_)(______) (__) (____)(_)\_) (__)
github.com/AnalogJ/scrutiny                            dev-0.4.14

Start the scrutiny server
time="2022-06-29T11:02:54-03:00" level=info msg="Trying to connect to scrutiny sqlite db: /opt/scrutiny/config/scrutiny.db\n"
[GIN-debug] [WARNING] Running in "debug" mode. Switch to "release" mode in production.
 - using env:   export GIN_MODE=release
 - using code:  gin.SetMode(gin.ReleaseMode)

time="2022-06-29T11:02:54-03:00" level=info msg="Successfully connected to scrutiny sqlite db: /opt/scrutiny/config/scrutiny.db\n"
time="2022-06-29T11:02:54-03:00" level=debug msg="InfluxDB url: http://localhost:8086"
time="2022-06-29T11:02:54-03:00" level=debug msg="Determine Influxdb setup status..."
time="2022-06-29T11:02:54-03:00" level=debug msg="Influxdb un-initialized, running first-time setup..."
time="2022-06-29T11:02:57-03:00" level=info msg="Database migration starting. Please wait, this process may take a long time...."

2022/06/29 11:02:57 /go/src/github.com/analogj/scrutiny/webapp/backend/pkg/database/scrutiny_repository_migrations.go:84
[...]

Issue #185, opened 2021-07-15
[BUG]


wallacebrf:
i have scrutiny running in docker on one of my synology NAS devices as both a web/API and collector. that is working fine and it is seeing all 9x of the drives on that device. 

i am trying to get the collector configured on a second synology NAS and i cannot get the collector to send the data to the web/API on the first device

here is the result of running the collector manally:

```
docker exec analogj-scrutiny-collector /scrutiny/bin/scrutiny-collector-metrics run
2021/07/15 17:22:48 Loading configuration file: /scrutiny/config/collector.yaml

 ___   ___  ____  __  __  ____  ____  _  _  _  _
/ __) / __)(  _ \(  )(  )(_  _)(_  _)( \( )( \/ )
\__ \( (__  )   / )(__)(   )(   _)(_  )  (  \  /
(___/ \___)(_)\_)(______) (__) (____)(_)\_) (__)
AnalogJ/scrutiny/metrics                               dev-0.3.12

time="2021-07-15T17:22:48Z" level=info msg="Verifying required tools" type=metrics
time="2021-07-15T17:22:48Z" level=info msg="Executing command: smartctl --scan -j" type=metrics
time="2021-07-15T17:22:49Z" level=info msg="Executing command: smartctl --info -j -d sat /dev/sdea" type=metrics
time="2021-07-15T17:22:49Z" level=error msg="Could not retrieve device information for sdea: exit status 2" type=metrics
time="2021-07-15T17:22:49Z" level=info msg="Executing command: smartctl --info -j -d sat /dev/sdec" type=metrics
time="2021-07-15T17:22:49Z" level=error msg="Could not retrieve device information for sdec: exit status 2" type=metrics
time="2021-07-15T17:22:49Z" level=info msg="Executing command: smartctl --info -j -d sat /dev/sded" type=metrics
time="2021-07-15T17:22:49Z" level=error msg="Could not retrieve device information for sded: exit status 2" type=metrics
time="2021-07-15T17:22:49Z" level=info msg="Executing command: smartctl --info -j -d sat /dev/sdee" type=metrics
time="2021-07-15T17:22:49Z" level=error msg="Could not retrieve device information for sdee: exit status 2" type=metrics
time="2021-07-15T17:22:49Z" level=info msg="Executing command: smartctl --info -j -d sat /dev/sda" type=metrics
time="2021-07-15T17:22:49Z" level=error msg="Could not retrieve device information for sda: exit status 2" type=metrics
[...]

Issue #183, opened 2021-06-20
[BUG] scrutiny-collector-metrics run ERROR: Post http://localhost:8080/api/devices/register


Davrizio:
I successfully started a docker and can access the webui. I went to run `scrutiny-collector-metrics run` and it is able to get metrics but has this error when trying to pass it to the api:

`ERROR: Post http://localhost:8080/api/devices/register: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)`

I do not have the  ` - SCRUTINY_API_ENDPOINT=` enabled

I am not sure where to look when trying to troubleshoot. Any help would be greatly appreciated!

Davrizio:
This seems to have sorted itself out. I restarted and updated the host machine and ran again. Everything showed in the webui after that!

PR #391, opened 2022-11-05
Increase collector timeout

adripo:
This pr increases http requests timeout.
Will fix old issues like #318 #185 #183

The problem happens when smartctl returns lot of data for one HDD and the http post request will not end in 10 seconds.
The error is: `Client.Timeout exceeded while awaiting headers`
I increased the timeout to 60 seconds and tested eveything. It solves the issue completely.

### DIFF
--- a/collector/pkg/collector/base.go
+++ b/collector/pkg/collector/base.go
@@ -8,7 +8,7 @@ import (
 	"time"
 )
 
-var httpClient = &http.Client{Timeout: 10 * time.Second}
+var httpClient = &http.Client{Timeout: 60 * time.Second}
 
 type BaseCollector struct {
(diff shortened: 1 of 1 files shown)


AnalogJ:
thanks for the PR! 🥳 

Wow, I never considered post payloads larger than 10s -- the SMART data has been reasonably sized in my experience, but I understand that network issues could cause these timeouts as well. 
I'm happy to merge this, however if you have an example of a payload that failed due to a timeout, I'd love to take a look at it (just send it to jason@thesparktree.com)

Thanks again!

AnalogJ:
Thanks for sending over your log files, I'll take a look and see if there's an unexpected cause for the slowness that I can determine.

PR #394, opened 2022-11-06
fixing bug. Null value for temperatures should be ignored.

### LINKED ISSUE #393: [BUG] Invalid temp values in database
**Describe the bug**
The temp values showed for one of my HDD is wrong. They are below 25C and this is not possible.
I debugged a little and found that the data in influxdb has invalid temp values of 0C. That changes the calculated median and generates wrong calculated values.

**Expected behavior**
Correct temp values

**Screenshots**
image
image

AnalogJ:
fixes #393

### DIFF
--- a/webapp/backend/pkg/database/scrutiny_repository_temperature.go
+++ b/webapp/backend/pkg/database/scrutiny_repository_temperature.go
@@ -17,6 +17,10 @@ func (sr *scrutinyRepository) SaveSmartTemperature(ctx context.Context, wwn stri
 	if len(collectorSmartData.AtaSctTemperatureHistory.Table) > 0 {
 
 		for ndx, temp := range collectorSmartData.AtaSctTemperatureHistory.Table {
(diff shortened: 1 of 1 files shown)


webapp/backend/pkg/database/scrutiny_repository_temperature.go
```diff
@@ -17,6 +17,10 @@ func (sr *scrutinyRepository) SaveSmartTemperature(ctx context.Context, wwn stri
 	if len(collectorSmartData.AtaSctTemperatureHistory.Table) > 0 {
 
 		for ndx, temp := range collectorSmartData.AtaSctTemperatureHistory.Table {
+			//temp value may be null, we must skip/ignore them. See #393
+			if temp == 0 {
```
adripo, line 21: I would suggest to use:
```
if temp != 0 {
    minutesOffset....
    ...
}
```

adripo (changes requested):

PR #396, opened 2022-11-10
Feat: dynamic line stroke settings

adripo:
New feature to configure line type in the graph of the temperatures:
image
image
image


AnalogJ:
Hey @adripo can you confirm that my changes worked for you?

PR #398, opened 2022-11-12
Angular refactoring to v13

adripo:
This is a full refactoring and cleanup of the frontend interface, keeping only the necessary dependencies:
- reduce build time
- build with docker
- increase efficiency
- better development support


AnalogJ:
Hey @adripo This is great, thanks! 🥳 

couple of quick notes. 

1. CI is still failing (I think) because the Makefile hardcodes a specific version of angular CLI.
2. The new database setting you added needs to be in its own DB migration, otherwise it won't be run for existing users. 
	- if you give me edit access to your PR/branch I can make the change for you and push this into master :)

Thanks again!

adripo:
Hey @AnalogJ, you already have write access to the branches associated with my PRs. About your suggestions:
1. I am still editing the Dockerfile before updating the CI actions, because I am refactoring the build process to complete the FE build inside the container. About this, is there any reason for using debian and not alpine in the runtime image?
2. I mistakenly included the line stroke settings in this PR, but there is a dedicated one here: #396 . Can you update that one with the migration script?

Thank you

AnalogJ:
apologies, I thought there was a checkbox or something visible on PRs where the maintainers could update the branch, but it looks like its just some simple text in the UI. 

1. I mostly stay with Debian rather than move to Alpine because I have a working container already. Getting Cron/smartmontools/timezone config all working correctly was a bit painful, so its easier to just stick with it. Are you considering migrating to Alpine to slim down the container or for security reasons? If you really want to migrate to something else, I'd be open to a version based on https://github.com/GoogleContainerTools/distroless -- better security and incredibly slim, though alot of work to implement. 

2. Submitted my migration change to that PR. https://github.com/AnalogJ/scrutiny/pull/396

AnalogJ:
hey @adripo have you had a chance to look into this? I'd love to merge this PR, we're on a pretty old version of Angular which I think is breaking our CI/Release builds

PR #406, opened 2022-11-28
Addition of notification testing command to troubleshooting

boomam:
Addition of notification testing command to troubleshooting.


AnalogJ:
thanks for the PR 🥳

PR #472, opened 2023-04-06
Allow configuration files with yml extension

### LINKED ISSUE #79: Support both .yml and .yaml for config file in default path.
As describled in #25
the service only looks for "/scrutiny/config/scrutiny.yaml" with the "yaml" extension. It would make a lot of sense to look for ".yml" as well since the short version of the extension is as common, if not more so, as ".yaml".

adamantike:
If a `collector.yml` or `scrutiny.yml` configuration file is present, use it as long as a `.yaml` version is not available too.

Fixes #79

### DIFF
--- a/collector/cmd/collector-metrics/collector-metrics.go
+++ b/collector/cmd/collector-metrics/collector-metrics.go
@@ -30,8 +30,14 @@ func main() {
 		os.Exit(1)
 	}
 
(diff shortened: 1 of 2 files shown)


AnalogJ:
Thanks for the PR @adamantike !
Haha this issue has been open for a while, thanks for picking it up.

Issue #510, opened 2023-09-01
[BUG] Observed Threshold Data for Smart 188 is incorrect


kaysond:
**Describe the bug**
The thresholds for Smart 188 are incorrect:
https://github.com/AnalogJ/scrutiny/blob/ee893cc360276aaf9684edb24c5fc0b61bd3a2e5/webapp/backend/pkg/thresholds/ata_attribute_metadata.go#L662-L725

If you look at the [backblaze data](https://www.backblaze.com/blog-smart-stats-2014-8.html#S188R), the thresholds should be 0-13 **billion**, 13-26 **billion**, and so on.

This error is probably what's causing confusion like #382. Also, given that the first interval is so ridiculously huge (because the highest interval is so high), I think it might make sense to deliberately change the AFR to be 0.09999, so it doesn't trigger the warning.

This issue actually caused me to go out and buy two new hard drives to replace what I thought was a failing drive in a RAID1. Only now that I see the replacements are also failing, despite being very new, did I do some research and find that they're actually completely fine.

Happy to submit a PR on this. Might be worth writing a script grabbing all of the latest data and updating every attribute.

kaysond:
closed by #512

Issue #382, opened 2022-10-13
[BUG] Failed - Command Timeout Segate


Susosvk:
**Describe the bug**
I have Segate HDD USB and debian OS - DIsable USB -UAS
I check hdd on windows smart check and disk is Healt OK

**Scree
![Untitled](https://user-images.githubusercontent.com/21984307/195530364-2fe5cfef-0116-4455-ba12-356922306cf4.png)
nshots**

ForsakenRei:
I saw similar things on two of my SATA SSDs, they both passed smart short test and show command timeout in Scrutiny. 

While if you use `smartctl -a /dev/sdb` you should be able to see whether it's really have command timeout, for my case yes they do. As the numbers are quite small and didn't change and they happened before I added them to my system, it's generally OK. It's a critical attributes for sure, if you see the number increases then it might need to be replaced.

Susosvk:
I use command smartctl -a /dev/sdb its all ok but dont show me ID 188
Segate dont have ID 188 on smartctl 

https://www.smartmontools.org/wiki/AttributesSeagate

Meaby fix for this bug - Segate hdd ignore ID 188

wanderling:
I am also running into this particular bug. Scrutiny marks my two brand new Seagate drives as failed with excessive 'Command Timeouts'. Running smartctl -a on the drives in question they pass all smart checks: 
<pre><code>SMART Attributes Data Structure revision number: 10
Vendor Specific SMART Attributes with Thresholds:
ID# ATTRIBUTE_NAME          FLAG     VALUE WORST THRESH TYPE      UPDATED  WHEN_FAILED RAW_VALUE
  1 Raw_Read_Error_Rate     0x000f   078   077   006    Pre-fail  Always       -       58897940
  3 Spin_Up_Time            0x0003   097   096   000    Pre-fail  Always       -       0
  4 Start_Stop_Count        0x0032   100   100   020    Old_age   Always       -       284
  5 Reallocated_Sector_Ct   0x0033   100   100   010    Pre-fail  Always       -       0
  7 Seek_Error_Rate         0x000f   100   253   045    Pre-fail  Always       -       715885
  9 Power_On_Hours          0x0032   099   099   000    Old_age   Always       -       878
 10 Spin_Retry_Count        0x0013   100   100   097    Pre-fail  Always       -       0
 12 Power_Cycle_Count       0x0032   100   100   020    Old_age   Always       -       2
[...]

PR #512, opened 2023-09-03
Fix Smart 188 thresholds

kaysond:
Addresses #510 and by extension #382


AnalogJ:
thanks for the PR! 🥳 
This is a great catch, I used a script to calculate the boundaries, and it must have missed the units or something.

PR #520, opened 2023-09-29
Update permissions setting in Dockerfile.web

kaysond:
This fixes issues with assets loading when you run as non root users


AnalogJ:
Thanks for the PR 🥳

PR #529, opened 2023-10-18
Update README.md Links to compose files

KaeTuuN:
Links were not working, so I replaced them with working ones.


AnalogJ:
Thanks for the PR! 🥳 

I just verified that those links work correctly on Github. Are they broken in your IDE or somewhere else?

KaeTuuN:
Yes, they do not work with PyCharm for me (Flatpak on Nobara 39). Neither with Visual Studio Code (Flatpak & local Installation) on the same System.

AnalogJ:
Ideally these would point to the local files, but I'm fine with changing to the github master branch versions.

PR #547, opened 2023-11-27
Add support for disabling repeat notifications if the values haven't changed

### LINKED ISSUE #364: Ignore UltraDMA CRC Error Count unless it is increasing?
Hi,

I'm running the latest image (ghcr.io/analogj/scrutiny:master-omnibus) in docker on Unraid. One of my disks had an issue a long time ago that was due to a bad cable and as a result, the "UltraDMA CRC Error Count" is elevated (87). Scrutiny reports this as a failed disk even though the value is not incrementing. Should this be reported as a failed disk when it's working fine and as far as I know, at no increased risk of failure? If so, is it possible to mark it as "accepted" and then monitor for the value incrementing?

Thanks for a great app btw.

### LINKED ISSUE #504: [FEAT] Add options to limit repeat notifications
**Is your feature request related to a problem? Please describe.**
I have my collectors running on a one hour cron job. One of my offsite disks has an issue, so I'm getting a replacement. In the meantime, though I'm getting a notification from scrutiny every hour, every time the collector sends data containing the issue.

**Describe the solution you'd like**
It would be nice to have a way of limiting repeated notifications. If it's exactly the same, I could reduce it via an option to, say, once a day. If something else goes wrong with the drive, or another drive goes bad, though, then I should still get a notification.

### LINKED ISSUE #514: [FEAT] Manually change thresholds
**Is your feature request related to a problem? Please describe.**
So I have a couple questionable disks that Scrutiny said they failed, but for now I'm taking the risk of using them.
Would be nice to say to Scrutiny that I accept the current situation as non-critical, so only flag them as failed if they get worse.

**Describe the solution you'd like**
I would like to manually increate the tresholds for a given disk.
Lets say on /dev/sda the Reallocated Sectors Count (0x05) is 8192 for this SSD, which is not good, but because it's not changing, I could set the config that 8192 is still WARN.

**Additional context**
Is the `limits` in example.scrutiny.yaml Features Coming Soon section going to implement something like this? Coudnt
https://github.com/AnalogJ/scrutiny/blob/master/example.scrutiny.yaml#L90

kaysond:
fixes #504 
fixes https://github.com/AnalogJ/scrutiny/issues/364
fixes #514
related #553

Still needs testing and some refactoring


webapp/backend/pkg/database/scrutiny_repository_device_smart_attributes.go
```diff
@@ -84,6 +85,73 @@ func (sr *scrutinyRepository) GetSmartAttributeHistory(ctx context.Context, wwn
...
+			fmt.Printf("Query error: %s\n", result.Err().Error())
+		}
+	} else {
+		return nil, err
+	}
+
+	//we have to sort the smartResults again, because the `union` command will return multiple 'tables' and only sort the records in each table.
+	sortSmartMeasurementsDesc(smartResults)
+
+	return smartResults, nil
```
kaysond: This is nearly identical to `GetSmartAttributeHistory()`, so it probably makes sense to refactor the commonalities out

webapp/backend/pkg/notify/notify.go
```diff
@@ -54,52 +56,79 @@ func ShouldNotify(device models.Device, smartAttrs measurements.Smart, statusThr
...
+						}
+					}
+				}
+				if old, ok := lastPoints[0].Attributes[attrId].(*measurements.SmartScsiAttribute); ok {
+					if current, ok := smartAttrs.Attributes[attrId].(*measurements.SmartScsiAttribute); ok {
+						if old.TransformedValue != current.TransformedValue {
+							return true
+						}
+					}
+				}
```
kaysond: There's probably a better way of doing this

kaysond:
@AnalogJ - would you mind reviewing before I update the tests?

webapp/backend/pkg/database/interface.go
```diff
@@ -20,7 +21,7 @@ type DeviceRepo interface {
 	DeleteDevice(ctx context.Context, wwn string) error
 
 	SaveSmartAttributes(ctx context.Context, wwn string, collectorSmartData collector.SmartInfo) (measurements.Smart, error)
-	GetSmartAttributeHistory(ctx context.Context, wwn string, durationKey string, attributes []string) ([]measurements.Smart, error)
+	GetSmartAttributeHistory(ctx context.Context, wwn string, durationKey string, n int, offset int, attributes []string) ([]measurements.Smart, error)
```
AnalogJ: I lean towards verbose variable names, so I don't love seeing `n`, how does `selectEntries` sound instead?

```suggestion
	GetSmartAttributeHistory(ctx context.Context, wwn string, durationKey string, selectEntries int, selectEntriesOffset int, attributes []string) ([]measurements.Smart, error)
```

webapp/backend/pkg/database/scrutiny_repository_device_smart_attributes.go
```diff
@@ -30,14 +31,14 @@ func (sr *scrutinyRepository) SaveSmartAttributes(ctx context.Context, wwn strin
 }
 
 // GetSmartAttributeHistory MUST return in sorted order, where newest entries are at the beginning of the list, and oldest are at the end.
-func (sr *scrutinyRepository) GetSmartAttributeHistory(ctx context.Context, wwn string, durationKey string, attributes []string) ([]measurements.Smart, error) {
+func (sr *scrutinyRepository) GetSmartAttributeHistory(ctx context.Context, wwn string, durationKey string, n int, offset int, attributes []string) ([]measurements.Smart, error) {
```
AnalogJ:
```suggestion
func (sr *scrutinyRepository) GetSmartAttributeHistory(ctx context.Context, wwn string, durationKey string, selectEntries int, selectEntriesOffset int, attributes []string) ([]measurements.Smart, error) {
```

webapp/backend/pkg/database/scrutiny_repository_device_smart_attributes.go
```diff
@@ -30,14 +31,14 @@ func (sr *scrutinyRepository) SaveSmartAttributes(ctx context.Context, wwn strin
...
 	// Get SMartResults from InfluxDB
 
 	//TODO: change the filter startrange to a real number.
 
 	// Get parser flux query result
 	//appConfig.GetString("web.influxdb.bucket")
-	queryStr := sr.aggregateSmartAttributesQuery(wwn, durationKey)
+	queryStr := sr.aggregateSmartAttributesQuery(wwn, durationKey, n, offset, attributes)
```
AnalogJ:
```suggestion
	queryStr := sr.aggregateSmartAttributesQuery(wwn, durationKey, selectEntries, selectEntriesOffset, attributes)
```

webapp/backend/pkg/database/scrutiny_repository_device_smart_attributes.go
```diff
@@ -99,7 +97,7 @@ func (sr *scrutinyRepository) saveDatapoint(influxWriteApi api.WriteAPIBlocking,
 	return influxWriteApi.WritePoint(ctx, p)
 }
 
-func (sr *scrutinyRepository) aggregateSmartAttributesQuery(wwn string, durationKey string) string {
+func (sr *scrutinyRepository) aggregateSmartAttributesQuery(wwn string, durationKey string, n int, offset int, attributes []string) string {
```
AnalogJ:
```suggestion
func (sr *scrutinyRepository) aggregateSmartAttributesQuery(wwn string, durationKey string, selectEntries int, selectEntriesOffset int, attributes []string) string {
```

webapp/backend/pkg/models/testdata/helper.go
```diff
@@ -32,7 +33,7 @@ func main() {
 		log.Fatalf("ERROR %v", err)
 	}
 	defer file.Close()
-	_, err = SendPostRequest("http://localhost:9090/api/devices/register", file)
+	_, err = SendPostRequest("http://localhost:8080/api/devices/register", file)
```
AnalogJ, line 36: Why change the port for testing?
on mac's 8080 is used by a built in service, so I use 9090 instead.

webapp/backend/pkg/database/scrutiny_repository_device_smart_attributes.go
```diff
@@ -140,34 +144,57 @@ func (sr *scrutinyRepository) aggregateSmartAttributesQuery(wwn string, duration
...
-			subQueryNames[0],
-			`|> yield()`,
-		}...)
-	} else {
-		partialQueryStr = append(partialQueryStr, []string{
-			fmt.Sprintf("union(tables: [%s])", strings.Join(subQueryNames, ", ")),
-			`|> sort(columns: ["_time"], desc: false)`,
-			`|> yield(name: "last")`,
-		}...)
+	return strings.Join(partialQueryStr, "\n")
```
AnalogJ: This looks correct, but I'd like to see some tests.

webapp/backend/pkg/notify/notify.go
```diff
@@ -54,52 +56,61 @@ func ShouldNotify(device models.Device, smartAttrs measurements.Smart, statusThr
...
 				if err != nil {
 					continue
 				}
-				if thresholds.AtaMetadata[attrIdInt].Critical {
-					hasFailingCriticalAttr = true
-				}
+				critical = thresholds.AtaMetadata[attrIdInt].Critical
+			}
+			if !critical {
+				continue
```
AnalogJ: Comments in this area would be helpful.

webapp/backend/pkg/database/scrutiny_repository_device_smart_attributes.go
AnalogJ: a comment would be helpful here. as well, describing how the selectEntries* keys are used -- and the fact that `selectEntriesOffset` is only relevant when `selectEntries` is >0. 

We could technically make `selectEntries` a `*int` type, to differentiate between 0 and "set", but that makes parameter passing kind-of annoying -- and it's unnecessary, since selecting 0 entries has no meaning.
I think we just need a big comment block here.

AnalogJ (changes requested):
Fantastic PR! 🥳  thanks for making such a meaningful contribution. 

I like the idea, and the implementation. Anything that touches the notification system needs alot of tests (which you mentioned you're working on). 
I also called out that I'd like to see some more comments in the filtering logic as well.

Thanks again!

webapp/backend/pkg/models/testdata/helper.go
kaysond: 8080 is the default port, right? It's also listed everywhere in the documentation. When I ran the helper, it didn't work initially because of the port mismatch.

I can revert this if you'd like, but it's probably a good idea to reconcile the number across the repo.

kaysond:
Addressed the comments and tested the image (it works). Will add/fix tests

kaysond:
@AnalogJ - I started looking into the tests, but I got stuck pretty quickly because I am a total golang noob.

The main issue is that I need to pass a `deviceRepo` to `ShouldNotify` now, which depends on a database that doesn't exist. I see some mocking libraries in other tests, so I'm guessing this might be the right approach? Would you mind pointing me in the direction of the right libraries/documentation/strategy to use?

Thanks!

kaysond:
Bump. Pinging @AnalogJ for some help with tests. Thanks!

AnalogJ:
I was going to say that it's super easy, but it seems like I forgot to generate a mock for the database. Let me generate one and push to your branch

AnalogJ:
So I added a new file: [webapp/backend/pkg/database/scrutiny_repository_device_smart_attributes.go](https://github.com/kaysond/scrutiny/blob/master/webapp/backend/pkg/database/scrutiny_repository_device_smart_attributes.go)

You can now use that to mock the database in your tests (similar to what I do with the Config mock):

``` go
	//setup
	mockCtrl := gomock.NewController(t)
	defer mockCtrl.Finish()
	fakeDatabase := mock_config.NewMockDeviceRepo(mockCtrl)

	// mock/fake out data and calls to the `fakeDatabase`, providing expected arguments and expected return parameters. 
    // you can mock any function in the interface: https://github.com/kaysond/scrutiny/blob/master/webapp/backend/pkg/database/interface.go
	fakeDatabase.EXPECT().GetDevices(context.Background()).Return([]models.Device{}, nil).AnyTimes()


    // if your inputs vary, you can use `gomock.Any()` as the argument
	fakeDatabase.EXPECT().HealthCheck(gomock.Any()).Return(fmt.Errorf("my error").AnyTimes()


	// once you finish setting up the mock, you can then pass it into your function under test:
	ShouldNotify(... , fakeDatabase) 
	
	// and internally when it calls `deviceRepo.GetSmartAttributeHistory(...)`, you can fake the response data. 
```

AnalogJ:
you should be good to write tests now @kaysond

kaysond:
> you should be good to write tests now @kaysond

Thank you! I'll take a crack at it tonight if I can.

kaysond:
@AnalogJ fixed the existing tests and added a few more for the case where repeat notifications are turned off. Looking at the updated `shouldNotify` code, this gives pretty good coverage of all the branching and looping.

kaysond:
Bump. Pinging @AnalogJ for review

AnalogJ (approved):

AnalogJ:
hey @kaysond apologies for the delay.
As far as I can tell everything looks good! thanks so much for pushing this through to the finish-line. 
I'm happy to merge it if you're ready.

kaysond:
> hey @kaysond apologies for the delay. As far as I can tell everything looks good! thanks so much for pushing this through to the finish-line. I'm happy to merge it if you're ready.

@AnalogJ please go ahead. And if you don't mind doing a release too then I can switch off my fork :)

AnalogJ:
will do!

PR #551, opened 2023-12-03
Update 188 attribute to cover 0 value

aivus:
Hello

I noticed that in case 188 attribute has value `0` it is shown as `warn` with the `Could not determine Observed Failure Rate for Critical Attribute` tooltip.

This is happening because [comparison logic in smart_ata_attribute.go](https://github.com/AnalogJ/scrutiny/blob/5bbd4c3b64f24139350e6d9c27548151a046d116/webapp/backend/pkg/models/measurements/smart_ata_attribute.go#L135-L137) doesn't include `low` value. But in case `low == high` bucket will be selected.


AnalogJ:
I think updating the threshold logic makes more sense than creating a 0,0 bucket. What do you think?

aivus:
I'll try to find time to rework logic

kaysond:
closing as stale. feel free to rebase and reopen

PR #557, opened 2023-12-18
Add option to discard SCT Data Table Temperature History (fix #494)

### LINKED ISSUE #494: [BUG] Temperature history showing incorrect numbers
**Describe the bug**
After running an initial collect, my temperature history graph looks like this:

image

And the drive's temperature history:

```
  "ata_sct_temperature_history": {
    "version": 2,
    "sampling_period_minutes": 3,
    "logging_interval_minutes": 59,
    "temperature": {
      "op_limit_min": 10,
      "op_limit_max": 40,
      "limit_min": 5,
      "limit_max": 60
    },
    "size": 128,
    "index": 10,
    "table": [
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      43,
      43,
      44,
      44,
      43,
      43,
      43,
      43,
      43,
      43,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      45,
      45,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      43,
      43,
      43,
      43,
      43,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      44,
      43,
      43,
      43,
      43,
      43,
      43,
      43,
      43,
      44,
      44,
      44,
      42,
      42,
      42,
      42,
      43,
      45,
      46,
      46,
      46
    ]
  },

```
I can assure you that the graph is displaying the temperature in reverse order. I've been utilizing the disk in the last few hours indicated by the rise in temperature, while the dashboard displays 46 as the current temperature as well.

mcarbonne:
~~This PR allow temperature history data to be retrieved correctly, using `Index` and `Size` fields.~~
~~`collectorSmartData.AtaSctTemperatureHistory.Table` is now managed as a circular buffer, starting from `Index`.~~

EDIT : I was wrong. The SCT temperature history table is indeed a circular buffer starting from index.
**But when outputting json**, `smartmontools` do re-order entries: first = oldest, last = most recent (see `ataPrintSCTTempHist` from `smartmontools`).

Details about spec here:
https://tc.gts3.org/cs3210/2016/spring/r/hardware/ATA8-ACS.pdf
(table 82 mostly)
And json generation from smartmontools here:
https://github.com/mirror/smartmontools/blob/master/ataprint.cpp (function `ataPrintSCTTempHist`)

Anyway, based on my observations, `ataPrintSCTTempHist` isn't reliable. I tested on 4 SSD (kingston, 2x micron, crucial) and none are fully compliant with the specification.

Here are some issues I found:
1. `logging_interval_minutes` / `sampling_period_minutes` isn't always correct.
Example (Micron 5200 Pro):
```
 286    2024-01-02 18:40    29  **********
 287    2024-01-02 18:41    30  ***********
 288    2024-01-02 18:42    29  **********
 ...    ..(  6 skipped).    ..  **********
 295    2024-01-02 18:49    29  **********
 296    2024-01-02 18:50    30  ***********
 ...    ..( 11 skipped).    ..  ***********
 308    2024-01-02 19:02    30  ***********
```
2 minutes later : 
```
 284    2024-01-02 18:38    29  **********
 285    2024-01-02 18:39    30  ***********
 286    2024-01-02 18:40    29  **********
 ...    ..(  6 skipped).    ..  **********
 293    2024-01-02 18:47    29  **********
 294    2024-01-02 18:48    30  ***********
 ...    ..( 15 skipped).    ..  ***********
 310    2024-01-02 19:04    30  ***********
```
Most likely, the "real" logging interval is ~ 30 seconds, but this interval is stored in an integer, it cannot work.
Therefore, only the current temperature (the last element of the table from json output) is correct.

2. When the drive starts (power on/resume), the drive is supposed to insert a special value (`0x80`, translated to `null` in json and `0` in `go`) but:
- some ssd (2 out of 4 I tested) do not insert the value `0x80` when resuming from sleep.
- none insert the magic value when powering on.

All values before this magical values must be ignored (because timestamp cannot be guessed). The current implementation only ignore "null" values, but not all the values before.

3. ~~Current implementation seems to parse the history the wrong way (first=most recent, last=oldest).~~

@AnalogJ maybe you have more data than me to confirm my observations. If you agree, my proposal is to simply remove history retrieval.

Note: This will fix #494 once for all

### DIFF
--- a/webapp/frontend/src/app/layout/common/dashboard-settings/dashboard-settings.component.html
+++ b/webapp/frontend/src/app/layout/common/dashboard-settings/dashboard-settings.component.html
@@ -102,6 +102,23 @@
                 </mat-select>
             </mat-form-field>
         </div>
+
+        <div class="mt-6 mb-2">
+            <h3 class="text-lg font-medium">Quirks</h3>
+            <div class="w-full border-b mt-1"></div>
+        </div>
+
+        <div class="flex flex-col mt-5 gt-md:flex-row">
+            <mat-form-field class="flex-auto gt-xs:pr-3 gt-md:pr-3"
+                matTooltip="Discard historical temperature data retrieved from the SMART Command Transport (SCT) Data Table, which may be inaccurate for some drives. The current temperature is always stored."
+                matTooltipPosition="right">
+                <mat-label>Discard SCT Temperature History</mat-label>
+                <mat-select [(ngModel)]=discardSCTTempHistory>
+                    <mat-option [value]=true>Enabled</mat-option>
+                    <mat-option [value]=false>Disabled</mat-option>
+                </mat-select>
+            </mat-form-field>
+        </div>
     </div>
 
 </mat-dialog-content>
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -163,7 +163,7 @@ docker cp scrutiny:/tmp/web.log web.log
 # Docker Development
 
 ```
-docker build -f docker/Dockerfile . -t chcr.io/analogj/scrutiny:master-omnibus
+docker build -f docker/Dockerfile . -t ghcr.io/analogj/scrutiny:master-omnibus
 docker run -it --rm -p 8080:8080 \
 -v /run/udev:/run/udev:ro \
 --cap-add SYS_RAWIO \
--- a/webapp/backend/pkg/database/interface.go
+++ b/webapp/backend/pkg/database/interface.go
@@ -26,7 +26,7 @@ type DeviceRepo interface {
 	SaveSmartAttributes(ctx context.Context, wwn string, collectorSmartData collector.SmartInfo) (measurements.Smart, error)
 	GetSmartAttributeHistory(ctx context.Context, wwn string, durationKey string, selectEntries int, selectEntriesOffset int, attributes []string) ([]measurements.Smart, error)
 
-	SaveSmartTemperature(ctx context.Context, wwn string, deviceProtocol string, collectorSmartData collector.SmartInfo) error
+	SaveSmartTemperature(ctx context.Context, wwn string, deviceProtocol string, collectorSmartData collector.SmartInfo, discardSCTTempHistory bool) error
 
 	GetSummary(ctx context.Context) (map[string]*models.DeviceSummary, error)
 	GetSmartTemperatureHistory(ctx context.Context, durationKey string) (map[string][]measurements.SmartTemperature, error)
--- a/webapp/backend/pkg/database/mock/mock_database.go
+++ b/webapp/backend/pkg/database/mock/mock_database.go
@@ -228,17 +228,17 @@ func (mr *MockDeviceRepoMockRecorder) SaveSmartAttributes(ctx, wwn, collectorSma
 }
 
 // SaveSmartTemperature mocks base method.
-func (m *MockDeviceRepo) SaveSmartTemperature(ctx context.Context, wwn, deviceProtocol string, collectorSmartData collector.SmartInfo) error {
+func (m *MockDeviceRepo) SaveSmartTemperature(ctx context.Context, wwn, deviceProtocol string, collectorSmartData collector.SmartInfo, discardSCTTempHistory bool) error {
 	m.ctrl.T.Helper()
-	ret := m.ctrl.Call(m, "SaveSmartTemperature", ctx, wwn, deviceProtocol, collectorSmartData)
+	ret := m.ctrl.Call(m, "SaveSmartTemperature", ctx, wwn, deviceProtocol, collectorSmartData, discardSCTTempHistory)
 	ret0, _ := ret[0].(error)
 	return ret0
 }
 
 // SaveSmartTemperature indicates an expected call of SaveSmartTemperature.
-func (mr *MockDeviceRepoMockRecorder) SaveSmartTemperature(ctx, wwn, deviceProtocol, collectorSmartData interface{}) *gomock.Call {
+func (mr *MockDeviceRepoMockRecorder) SaveSmartTemperature(ctx, wwn, deviceProtocol, collectorSmartData, discardSCTTempHistory interface{}) *gomock.Call {
 	mr.mock.ctrl.T.Helper()
-	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "SaveSmartTemperature", reflect.TypeOf((*MockDeviceRepo)(nil).SaveSmartTemperature), ctx, wwn, deviceProtocol, collectorSmartData)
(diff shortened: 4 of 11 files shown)


mcarbonne:
Also, to mitigate this change, the default `COLLECTOR_CRON_SCHEDULE` might be reduced from one day to one hour (`COLLECTOR_CRON_SCHEDULE=0 * * * *`).

mcarbonne:
@AnalogJ I'd like to have your feedback on this PR.
My proposal was to simply remove temperature history retrieval because of its lack of robustness (cf. my arguments above) but If you don't agree, I can update my PR to add an option to disable this feature instead of removing it.

Here is a screenshot of my temperature history graph (with temperature history retrieval enabled) :
image
Obviously, there is something wrong with the temperature history of the green one.

AnalogJ:
Hey @mcarbonne appreciate your work here -- but yeah I'd rather add a switch to enable/disable it, rather than remove it completely

mcarbonne:
> Hey @mcarbonne appreciate your work here -- but yeah I'd rather add a switch to enable/disable it, rather than remove it completely

Thanks for you feedback. I have updated my PR with your suggestions but I have not yet been able to test this change: I have build issues (npm related) but haven't found out if it's related to my changes but it is similar to the nightly build issue. I'll try to investigate when I have time.

mcarbonne:
@AnalogJ build is working again, feature is behaving as expected, docker image is available (ghcr.io/mcarbonne/scrutiny:master-omnibus). Feel free to give me your feedback.

mcarbonne:
@AnalogJ PR updated :)

helgehatt:
Would be great if this could get merged anytime soon.

AnalogJ:
This looks good, let me check why the CI isn't running.

mcarbonne:
I did some testing on my fork and indeed, workflows need a little update. I included it in my PR (commit b7d57c03d5a2752b98b493d9c21b2ec261b39da1).

This is mandatory according to documentation:
https://github.com/actions/upload-artifact


> In matrix scenarios, be careful to not accidentally upload to the same artifact, or else you will encounter conflict errors. It would be best to name the artifact with a prefix or suffix from the matrix:
> 
> jobs:
>   upload:
>     name: Generate Build Artifacts
> 
>     strategy:
>       matrix:
>         os: [ubuntu-latest, windows-latest]
>         version: [a, b, c]
> 
>     runs-on: ${{ matrix.os }}
> 
>     steps:
>     - name: Build
>       run: ./some-script --version=${{ matrix.version }} > my-binary
>     - name: Upload
>       uses: actions/upload-artifact@v4
>       with:
>         name: binary-${{ matrix.os }}-${{ matrix.version }}
>         path: my-binary
> 
> This will result in artifacts like: binary-ubuntu-latest-a, binary-windows-latest-b, and so on.

EDIT : this is a breaking change with v4: `Unlike earlier versions of upload-artifact, uploading to the same artifact via multiple jobs is not supported with v4.`

ssorgatem:
so how's the merging of this looking?

mcarbonne:
@kaysond I've cleaned the history of my PR ! (rebase and squash).
Let me know if you agree with my changes and/or if further changes are needed.
Note : this feature has been tested and running for a while in my fork and also several other forks.

kaysond:
> @kaysond I've cleaned the history of my PR ! (rebase and squash). Let me know if you agree with my changes and/or if further changes are needed. Note : this feature has been tested and running for a while in my fork and also several other forks.

Thanks! Took a quick look from my phone and don't have any immediate concerns. I do think it would be nice to have this be a per-drive setting, but obviously things aren't set up to accommodate that.

My only thought is to rename the config option to something more user-friendly. Without documentation it may not be obvious to people that it doesn't fully disable temp logging...

mcarbonne:
I added a tooltip + a dedicated section for this kind of options :
tooltip

I agree that a per device option would be better but the current software structure is limited right now.

webapp/frontend/src/app/layout/common/dashboard-settings/dashboard-settings.component.html
```diff
@@ -102,6 +102,23 @@ <h2 mat-dialog-title>Scrutiny Settings</h2>
...
         </div>
+
+        <div class="mt-6 mb-2">
+            <h3 class="text-lg font-medium">Quirks</h3>
+            <div class="w-full border-b mt-1"></div>
+        </div>
+
+        <div class="flex flex-col mt-5 gt-md:flex-row">
+            <mat-form-field class="flex-auto gt-xs:pr-3 gt-md:pr-3"
+                matTooltip="Retrieve historical temperature data from the drive's SCT log. This feature might be bugged on some drives."
```
kaysond:
```suggestion
                matTooltip="Discard historical temperature data retrieved from the SMART Command Transport (SCT) Data Table, which may be inaccurate for some drives. The current temperature is always stored."
```

kaysond (changes requested):
Looks good. Tooltip and quirks section is a good idea.

I know it's pedantic, but could you please update the variable names, UI, etc to use "discard" instead of "retrieve"? (Note that this also means the default should be FALSE). I think it's more clear and more accurate since the data is technically always retrieved. I also prefer that it's something that has to be enabled rather than disabled, which I think makes more sense for all quirks...

I'm open to a different word if you can think of a better one...

webapp/frontend/src/app/layout/common/dashboard-settings/dashboard-settings.component.html
mcarbonne: I updated my PR with your suggestion.

~~I also noticed something strange, not related to my changes, in the tests: some tests seems to always pass.
I did a quick POC in another branch on my fork : https://github.com/mcarbonne/scrutiny/commit/0f0e7082063a85644b60f7d77fb250562c45993b~~
```
	fakeConfig.EXPECT().GetBool("user.collector.discard_sct_temp_history").Return(false).AnyTimes()
	fakeConfig.EXPECT().GetBool("user.collector.discard_sct_temp_history").Return(true).AnyTimes()
```
~~But the pipeline is passing on this branch :upside_down_face:~~

EDIT: I did a little digging; this isn't a bug, but the test itself. I'm not very familiar with gomock, but it seems AnyTimes() allows the test to pass even if the function is never called.

kaysond (approved):
LGTM! Thanks for the updates and for your contribution. I'm going to do a sanity check deployment later today then will merge this.

PR #577, opened 2024-02-17
Update SUPPORTED_NAS_OS.md

DrFrankensteinUK:
Added another link for the new Container Manager version of my guide for those on the newer DSM versions. The older guide while archived still functions correctly.


AnalogJ:
thanks for the PR! 🥳

PR #585, opened 2024-02-26
chore: use Linux container CPU quota

### LINKED ISSUE #584: [BUG] Scrutiny should be cgroup aware
**Describe the bug**

[Go is not cgroup aware](https://github.com/golang/go/issues/33803). This means that Scrutiny will be throttled hard in containerised environments with CPU limits.

The below screenshot shows the container with a CPU limit in Kubernetes. The first half of the time series is on a 13600k, with 20 threads. The second half is when the machine changed to use an EPYC 7763 with 128 threads.

image

**Expected behavior**

Scrutiny should not be CPU throttling with adequate resources, and the throttling should not scale with the number of available threads.

[automaxprocs](https://github.com/uber-go/automaxprocs) can use cgroups automatically.

**Screenshots**

See above.

**Log Files**

N/A

uhthomas:
Go is not cgroup aware and by default will set GOMAXPROCS to the number of available threads, regardless of whether it is within the allocated quota. This behaviour causes high amount of CPU throttling and degraded application performance.

Fixes: #584

### DIFF
--- a/collector/cmd/collector-metrics/collector-metrics.go
+++ b/collector/cmd/collector-metrics/collector-metrics.go
@@ -17,6 +17,7 @@ import (
 	utils "github.com/analogj/go-util/utils"
 	"github.com/fatih/color"
 	"github.com/urfave/cli/v2"
+	_ "go.uber.org/automaxprocs"
 )
 
 var goos string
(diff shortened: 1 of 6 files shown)


uhthomas:
I don't really understand what codecov is unhappy about - this commit is rebased on master.

kaysond:
@uhthomas - can you rebase this?

uhthomas:
This is no longer needed, Go 1.25 does this by default.

kaysond:
> This is no longer needed, Go 1.25 does this by default.

@uhthomas - we're currently using 1.20.x across the board. if you'd like to help update it, that would be appreciated. https://github.com/AnalogJ/scrutiny/issues/872

PR #586, opened 2024-02-26
chore: gofumpt

uhthomas:
Format all Go files with gofumpt.

https://github.com/mvdan/gofumpt

	gofumpt -extra -w .


AnalogJ:
I'm not 100% convinced this is required. Can you give me some addl context @uhthomas ?

Issue #538, opened 2023-11-01
[BUG] Could not retrieve device information


arthurcruz1:
Drive is not shown on the WebUI

`docker compose exec scrutiny /opt/scrutiny/bin/scrutiny-collector-metrics run`

```
Error:
Could not retrieve device information for sda: exit status 2  type=metrics
```

`smartctl --info --json --device sat /dev/sda  type=metrics`

```
smartctl --info --json --device sat /dev/sda  type=metrics
{
  "json_format_version": [
    1,
    0
  ],
  "smartctl": {
    "version": [
      7,
      3
    ],
    "svn_revision": "5338",
    "platform_info": "aarch64-linux-5.15.0-1042-raspi",
    "build_info": "(local build)",
    "argv": [
      "smartctl",
      "--info",
      "--json",
      "--device",
      "sat",
      "/dev/sda",
      "type=metrics"
    ],
    "messages": [
      {
        "string": "ERROR: smartctl takes ONE device name as the final command-line argument.",
        "severity": "error"
      }
    ],
    "exit_status": 1
  }
}
```

`docker logs scrutiny`

```
 ___   ___  ____  __  __  ____  ____  _  _  _  _
/ __) / __)(  _ \(  )(  )(_  _)(_  _)( \( )( \/ )
\__ \( (__  )   / )(__)(   )(   _)(_  )  (  \  /
(___/ \___)(_)\_)(______) (__) (____)(_)\_) (__)
AnalogJ/scrutiny/metrics                                dev-0.7.2

time="2023-11-01T14:38:46Z" level=info msg="Verifying required tools" type=metrics
time="2023-11-01T14:38:46Z" level=info msg="Executing command: smartctl --scan --json" type=metrics
time="2023-11-01T14:38:46Z" level=info msg="Executing command: smartctl --info --json --device sat /dev/sda" type=metrics
time="2023-11-01T14:38:46Z" level=error msg="Could not retrieve device information for sda: exit status 2" type=metrics
time="2023-11-01T14:38:46Z" level=info msg="Sending detected devices to API, for filtering & validation" type=metrics
[...]

PR #596, opened 2024-03-09
rebase docker image to debian 12 (bookworm)

dropsignal:
docker image is now based on debian 12 bookworm so smartmontools is now 7.3.
to resolve [BUG] Could not retrieve device information #538 
https://github.com/AnalogJ/scrutiny/issues/538

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -18,7 +18,8 @@ WORKDIR /go/src/github.com/analogj/scrutiny
 COPY --link . /go/src/github.com/analogj/scrutiny
 RUN apt-get update && DEBIAN_FRONTEND=noninteractive \
     apt-get install -y --no-install-recommends \
-    file
+    file \
+    && rm -rf /var/lib/apt/lists/*
 RUN make binary-clean binary-all WEB_BINARY_NAME=scrutiny
 
 
@@ -40,15 +41,16 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive \
     curl \
     smartmontools \
     tzdata \
-    && update-ca-certificates
-RUN case ${TARGETARCH} in \
-        "amd64")  S6_ARCH=amd64  ;; \
-        "arm64")  S6_ARCH=aarch64  ;; \
-    esac \
+    && rm -rf /var/lib/apt/lists/* \
+    && update-ca-certificates \
+    && case ${TARGETARCH} in \
+           "amd64")  S6_ARCH=amd64  ;; \
+           "arm64")  S6_ARCH=aarch64  ;; \
+       esac \
     && curl https://github.com/just-containers/s6-overlay/releases/download/v${S6VER}/s6-overlay-${S6_ARCH}.tar.gz -L -s --output /tmp/s6-overlay-${S6_ARCH}.tar.gz \
     && tar xzf /tmp/s6-overlay-${S6_ARCH}.tar.gz -C / \
-    && rm -rf /tmp/s6-overlay-${S6_ARCH}.tar.gz
-RUN curl -L https://dl.influxdata.com/influxdb/releases/influxdb2-${INFLUXVER}-${TARGETARCH}.deb --output /tmp/influxdb2-${INFLUXVER}-${TARGETARCH}.deb \
+    && rm -rf /tmp/s6-overlay-${S6_ARCH}.tar.gz \
+    curl -L https://dl.influxdata.com/influxdb/releases/influxdb2-${INFLUXVER}-${TARGETARCH}.deb --output /tmp/influxdb2-${INFLUXVER}-${TARGETARCH}.deb \
     && ln -s /usr/bin/false /bin/false \
     && ln -s /usr/bin/bash /bin/bash \
     && dpkg -i --force-all /tmp/influxdb2-${INFLUXVER}-${TARGETARCH}.deb \
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -50,7 +50,7 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive \
     && curl https://github.com/just-containers/s6-overlay/releases/download/v${S6VER}/s6-overlay-${S6_ARCH}.tar.gz -L -s --output /tmp/s6-overlay-${S6_ARCH}.tar.gz \
     && tar xzf /tmp/s6-overlay-${S6_ARCH}.tar.gz -C / \
     && rm -rf /tmp/s6-overlay-${S6_ARCH}.tar.gz \
-    curl -L https://dl.influxdata.com/influxdb/releases/influxdb2-${INFLUXVER}-${TARGETARCH}.deb --output /tmp/influxdb2-${INFLUXVER}-${TARGETARCH}.deb \
+    && curl -L https://dl.influxdata.com/influxdb/releases/influxdb2-${INFLUXVER}-${TARGETARCH}.deb --output /tmp/influxdb2-${INFLUXVER}-${TARGETARCH}.deb \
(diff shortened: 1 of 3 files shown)


docker/Dockerfile
```diff
@@ -36,17 +40,20 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive \
     curl \
     smartmontools \
     tzdata \
-    && update-ca-certificates \
-    &&  case ${TARGETARCH} in \
-            "amd64")  S6_ARCH=amd64  ;; \
-            "arm64")  S6_ARCH=aarch64  ;; \
-        esac \
+    && update-ca-certificates
+RUN case ${TARGETARCH} in \
```
AnalogJ: is there a reason you wanted to split up this RUN into multiple steps? 
Is this a leftover from debugging?

docker/Dockerfile
```diff
@@ -12,22 +12,26 @@ RUN make binary-frontend
...
-FROM debian:bullseye-slim as runtime
+FROM debian:bookworm-slim as runtime
 ARG TARGETARCH
 EXPOSE 8080
 WORKDIR /opt/scrutiny
 ENV PATH="/opt/scrutiny/bin:${PATH}"
 ENV INFLUXD_CONFIG_PATH=/opt/scrutiny/influxdb
 ENV S6VER="1.21.8.0"
 ENV INFLUXVER="2.2.0"
+SHELL ["/usr/bin/sh", "-c"]
```
AnalogJ, line 35: Is this necessary?

docker/Dockerfile
```diff
@@ -36,17 +40,20 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive \
...
+        "arm64")  S6_ARCH=aarch64  ;; \
+    esac \
     && curl https://github.com/just-containers/s6-overlay/releases/download/v${S6VER}/s6-overlay-${S6_ARCH}.tar.gz -L -s --output /tmp/s6-overlay-${S6_ARCH}.tar.gz \
     && tar xzf /tmp/s6-overlay-${S6_ARCH}.tar.gz -C / \
-    && rm -rf /tmp/s6-overlay-${S6_ARCH}.tar.gz \
-    && curl -L https://dl.influxdata.com/influxdb/releases/influxdb2-${INFLUXVER}-${TARGETARCH}.deb --output /tmp/influxdb2-${INFLUXVER}-${TARGETARCH}.deb \
+    && rm -rf /tmp/s6-overlay-${S6_ARCH}.tar.gz
+RUN curl -L https://dl.influxdata.com/influxdb/releases/influxdb2-${INFLUXVER}-${TARGETARCH}.deb --output /tmp/influxdb2-${INFLUXVER}-${TARGETARCH}.deb \
+    && ln -s /usr/bin/false /bin/false \
+    && ln -s /usr/bin/bash /bin/bash \
```
AnalogJ, line 55: what's the intention here?

docker/Dockerfile
```diff
@@ -36,17 +40,20 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive \
...
     && dpkg -i --force-all /tmp/influxdb2-${INFLUXVER}-${TARGETARCH}.deb \
-    && rm -rf /tmp/influxdb2-2.2.0-${TARGETARCH}.deb
+    && rm -f /bin/bash \
+    && rm -rf /tmp/influxdb2-${INFLUXVER}-${TARGETARCH}.deb
```
AnalogJ, line 58: good catch! 👍

docker/Dockerfile.collector
```diff
@@ -4,16 +4,17 @@
...
 
 ########
-FROM golang:1.20-bullseye as backendbuild
+FROM golang:1.20-bookworm as backendbuild
 
 WORKDIR /go/src/github.com/analogj/scrutiny
 
 COPY . /go/src/github.com/analogj/scrutiny
 
+RUN apt-get update && apt-get install -y file
```
AnalogJ: what is `file` used for?

AnalogJ: we should cleanup the `apt/lists` after this package is installed:

https://docs.docker.com/develop/develop-images/instructions/#apt-get

AnalogJ (changes requested):
thanks for the PR! 🥳 
made a couple of comments -- mostly just questions

docker/Dockerfile
dropsignal: Yes. I was trying to figure out if it was the apt command or curl command causing the problem. It can be combined into a single line again.

docker/Dockerfile
dropsignal: Apparently so. When I first tried to rebase to bookworm, the build would generate errors about missing /bin/sh. According to https://www.debian.org/releases/bookworm/amd64/release-notes/ch-information.en.html#a-merged-usr-is-now-required, /bin no longer holds common comnands like /bin/sh (although they should still work). Changing the shell in the docker file seems to solve the issue.

docker/Dockerfile
dropsignal: This is related to the merged /usr required in bookworm. /bin/false doesn't exist so when the user is created the useradd command throws a warning. The simple solution was to create a symbolic link, create the user, the delete the symbolic link. No more warning is thrown.

docker/Dockerfile.collector
dropsignal: During make, file seems to be used to identify the binary. For what purpose, I don't know. To get rid of the error though I install file. It can probably be removed afterwards.

dropsignal:
I can make the necessary changes, but would that require a new pull request?

AnalogJ:
Nope, no new pull request required. Just make changes in your existing branch, the PR will be updated.

dropsignal (commented):
Requested changes were made based on previous comments. Specifically, recombining split RUN command from Dockerfile and adding apt list clean up as per Docker imaging building best practices.

docker/Dockerfile
```diff
@@ -36,17 +41,21 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive \
...
-        esac \
+    && case ${TARGETARCH} in \
+           "amd64")  S6_ARCH=amd64  ;; \
+           "arm64")  S6_ARCH=aarch64  ;; \
+       esac \
     && curl https://github.com/just-containers/s6-overlay/releases/download/v${S6VER}/s6-overlay-${S6_ARCH}.tar.gz -L -s --output /tmp/s6-overlay-${S6_ARCH}.tar.gz \
     && tar xzf /tmp/s6-overlay-${S6_ARCH}.tar.gz -C / \
     && rm -rf /tmp/s6-overlay-${S6_ARCH}.tar.gz \
-    && curl -L https://dl.influxdata.com/influxdb/releases/influxdb2-${INFLUXVER}-${TARGETARCH}.deb --output /tmp/influxdb2-${INFLUXVER}-${TARGETARCH}.deb \
+    curl -L https://dl.influxdata.com/influxdb/releases/influxdb2-${INFLUXVER}-${TARGETARCH}.deb --output /tmp/influxdb2-${INFLUXVER}-${TARGETARCH}.deb \
```
AnalogJ: missing `&&` before curl.

You should build this docker image locally and confirm it works.

AnalogJ (changes requested):
broken `curl` command -- but other than that, everything else looks good.

docker/Dockerfile
dropsignal: fixed. i rebuilt the docker image and it seems to work correctly. i verified that bookworm was used. i have a Vagrantfile that spins up one VM to build the docker image and then spins up another VM to actually run the build docker image. i can post that in another PR if you're interested.

AnalogJ:
thanks!

Issue #221, opened 2022-03-31
[Standby-Support][FEAT] Add support for `smartctl --nocheck=` to the collector


danie1k:
The `smartctl` tool has a bunch of useful arguments, Scrutiny's collector could make use of them, [similarly to Telegraf](https://github.com/influxdata/telegraf/tree/master/plugins/inputs/smart).  
I suggested only the `nocheck` argument, because for me personally it's the most important one because of "standby" option, so the drive is not woken up constantly.

**Describe the solution you'd like**
Extend collector's config file with more options, covering smartctl's arguments.

**Excerpt from smartctl manual**

> -n POWERMODE, --nocheck=POWERMODE
>
> [ATA only] Specifies if smartctl should exit before performing any checks when the device is in a low-power mode. It may be used to prevent a disk from being spun-up by smartctl. The power mode is ignored by default. A nonzero exit status is returned if the device is in one of the specified low-power modes (see RETURN VALUES below).
> 
> Note: If this option is used it may also be necessary to specify the device type with the '-d' option. Otherwise the device may spin up due to commands issued during device type autodetection.
> 
> The valid arguments to this option are:
> 
> never - check the device always, but print the power mode if '-i' is specified.
> 
> sleep - check the device unless it is in SLEEP mode.
> 
> standby - check the device unless it is in SLEEP or STANDBY mode. In these modes most disks are not spinning, so if you want to prevent a disk from spinning up, this is probably what you want.
> 
> idle - check the device unless it is in SLEEP, STANDBY or IDLE mode. In the IDLE state, most disks are still spinning, so this is probably not what you want. 

-- https://linux.die.net/man/8/smartctl

danie1k:
After getting more familiar with the repo... Will that be supported by the `command` in following section of the config file?

https://github.com/AnalogJ/scrutiny/blob/060ac7b83a7525c99bf2f03c26dcb3d89dafc976/example.scrutiny.yaml#L85-L88

AnalogJ:
Yeah, I eventually want to make the collector command flags completely customizable. 
I'll add this as an enhancement, but it'll have to wait until this database upgrade is complete.

AnalogJ:
apologies, I should have updated & closed this issue earlier. This functionality has been available since `v0.4.9`

You can update the collector config files and specify the command flags you'd like to use when Scrutiny triggers `smartctl`. 


```yaml
# example to show how to override the smartctl command args globally
commands:
[...]

PR #619, opened 2024-03-30
feat: Ability to override commands args

datenzar:
In order to override the arguments which are used e.g. to call smartctl, they need to be bind to the respective environment variable.

### DIFF
--- a/collector/pkg/config/config.go
+++ b/collector/pkg/config/config.go
@@ -48,15 +48,10 @@ func (c *configuration) Init() error {
 	c.SetDefault("commands.metrics_info_args", "--info --json")
 	c.SetDefault("commands.metrics_smart_args", "--xall --json")
 
-	// Bind environment variables with COLLECTOR_ prefix
-	c.BindEnv("host.id", "COLLECTOR_HOST_ID")
-	c.BindEnv("log.level", "COLLECTOR_LOG_LEVEL")
-	c.BindEnv("log.file", "COLLECTOR_LOG_FILE")
-	c.BindEnv("api.endpoint", "COLLECTOR_API_ENDPOINT")
-	c.BindEnv("commands.metrics_smartctl_bin", "COLLECTOR_COMMANDS_METRICS_SMARTCTL_BIN")
-	c.BindEnv("commands.metrics_scan_args", "COLLECTOR_COMMANDS_METRICS_SCAN_ARGS")
-	c.BindEnv("commands.metrics_info_args", "COLLECTOR_COMMANDS_METRICS_INFO_ARGS") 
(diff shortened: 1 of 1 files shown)


datenzar:
Hopefully this will fix https://github.com/AnalogJ/scrutiny/issues/221

collector/pkg/config/config.go
```diff
@@ -48,6 +48,16 @@ func (c *configuration) Init() error {
...
+	// Bind environment variables with COLLECTOR_ prefix
+	c.BindEnv("host.id", "COLLECTOR_HOST_ID")
+	c.BindEnv("log.level", "COLLECTOR_LOG_LEVEL")
+	c.BindEnv("log.file", "COLLECTOR_LOG_FILE")
+	c.BindEnv("api.endpoint", "COLLECTOR_API_ENDPOINT")
+	c.BindEnv("commands.metrics_smartctl_bin", "COLLECTOR_COMMANDS_METRICS_SMARTCTL_BIN")
+	c.BindEnv("commands.metrics_scan_args", "COLLECTOR_COMMANDS_METRICS_SCAN_ARGS")
+	c.BindEnv("commands.metrics_info_args", "COLLECTOR_COMMANDS_METRICS_INFO_ARGS") 
+	c.BindEnv("commands.metrics_smart_args", "COLLECTOR_COMMANDS_METRICS_SMART_ARGS")
+
```
AnalogJ:
```suggestion
	// Bind environment variables with COLLECTOR_ prefix
	c.SetEnvPrefix("COLLECTOR")
```

I think that should work instead. Can you test and confirm @datenzar ?

datenzar: Hi @AnalogJ 
sorry for my late reply, but I haven't had the chance to look into this issue for a while.

From what I understood by reading in the Viper documentation, the prefix alone isn't sufficient. 

The following two lines however should do the job
```go
c.SetEnvPrefix("COLLECTOR")
c.AutomaticEnv()
```

I tried the above lines, but the values weren't picked up. I'm not sure if I misunderstood the concept but following call (on Linux) should lead to an error due to an arbitrary non-existing command being used:
`COLLECTOR_COMMANDS_METRICS_SMARTCTL_BIN=non_existing_cmd go run collector/cmd/collector-metrics/collector-metrics.go run --config collector.yaml`

Yet the application won't pick up the value from env. BTW: It works perfectly, when removing the hierarchy (-> replace all keys `commands.metrics_smartctl_bin` with `commands_metrics_smartctl_bin`).

datenzar: NVM, I was able to fix it with a code piece I stumbled on your webapp backend. I was able to test it with success 🥳

PR #625, opened 2024-04-05
fixing cron in #602

### LINKED ISSUE #602: [BUG] Cron Schedule not working in v0.8.0
**Describe the bug**
Cron Schedule stop working after upgrade to v0.8.0

**Expected behavior**
Cron Schedule should executed as normal
AnalogJ/scrutiny/metrics  doesn't start as planned by cron

AnalogJ:
Updated s6overlay to v3
Note:  xz-utils was added as a requirement for s6-overlay (using safe 5.4.1 instead of compromised 5.6.x versions)
fixes #602

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -30,8 +30,9 @@ EXPOSE 8080
 WORKDIR /opt/scrutiny
 ENV PATH="/opt/scrutiny/bin:${PATH}"
 ENV INFLUXD_CONFIG_PATH=/opt/scrutiny/influxdb
-ENV S6VER="1.21.8.0"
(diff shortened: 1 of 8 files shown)


docker/Dockerfile
```diff
@@ -41,20 +42,22 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive \
     curl \
     smartmontools \
     tzdata \
+    procps \
+    xz-utils \
```
AnalogJ, line 46: note about `xz-utils`: I'm aware of the vulnerability - the compromised version was removed from the Debian upstream, and is not present in this image.

Issue #628, opened 2024-04-13
It would be nice if we could get the interface to show hours instead of years


DawnTreader:
I am an IT guy, I always look at the amount of hours rather than the nice human years

It would be nice if there was a way in the settings to change that output in the dashboard.

adprom:
Having both displayed would be ideal. As well as for "failed drives", mouseover showing the failed metric

bauzer714:
@adprom @DawnTreader this was merged and should be available in docker images.

AnalogJ:
closing now that the PR has been merged.

PR #634, opened 2024-05-08
Create a setting for user to indicate humanized or hours on dashboard/device detail

bauzer714:
Default is "humanize" creates a non-breaking change.

**Setting:**
image

image

Dashboard:
image

image

Detail:
image

Detail tooltip:
image

image

Implements original request of #628.

### DIFF
--- a/webapp/backend/pkg/database/scrutiny_repository_migrations.go
+++ b/webapp/backend/pkg/database/scrutiny_repository_migrations.go
@@ -333,13 +333,6 @@ func (sr *scrutinyRepository) Migrate(ctx context.Context) error {
 						SettingValueString:    "smooth",
 					},
 					{
-						SettingKeyName:        "powered_on_hours_unit",
-						SettingKeyDescription: "Presentation format for device powered on time ('humanize' | 'device_hours')",
-						SettingDataType:       "string",
-						SettingValueString:    "humanize",
-					},
-
-					{
 						SettingKeyName:        "metrics.notify_level",
 						SettingKeyDescription: "Determines which device status will cause a notification (fail or warn)",
 						SettingDataType:       "numeric",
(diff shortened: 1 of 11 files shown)


webapp/backend/pkg/database/scrutiny_repository_migrations.go
```diff
@@ -332,6 +332,12 @@ func (sr *scrutinyRepository) Migrate(ctx context.Context) error {
 						SettingDataType:       "string",
 						SettingValueString:    "smooth",
 					},
+					{
+						SettingKeyName:        "powered_on_hours_unit",
+						SettingKeyDescription: "Presentation format for device powered on time ('humanize' | 'device_hours')",
+						SettingDataType:       "string",
+						SettingValueString:    "humanize",
+					},
```
AnalogJ: This has to be added as a *new* migration, otherwise it will be skipped for users with an existing database..

AnalogJ:
sure, I can get behind this. 
thanks for the PR! 🥳 

Added comment about the db settings change - it needs to be in it's own migration

bauzer714:
Thanks @AnalogJ - I've never developed in Go, any specific naming conventions or versioning? Does it need to be removed from the current script?

AnalogJ:
Basically just copy paste a migration entry like this: https://github.com/bauzer714/scrutiny/blob/addDeviceHoursSetting/webapp/backend/pkg/database/scrutiny_repository_migrations.go#L379C3-L393C5

The ID should start with "m" and be followed by a timestamp in YYYYMMDDHHMMSS format

`date '+%Y%m%d%H%M%S'` should do it

Thanks @bauzer714 !

bauzer714:
@AnalogJ - thanks for the guidance. Pushed the latest changes. Let me know if this needs further changes.

webapp/frontend/src/app/shared/device-hours.pipe.spec.ts
```diff
@@ -0,0 +1,9 @@
+import { DeviceHoursPipe } from './device-hours.pipe';
+
+
+describe('DeviceHoursPipe', () => {
+  it('create an instance', () => {
+    const pipe = new DeviceHoursPipe();
+    expect(pipe).toBeTruthy();
+  });
+});
```
AnalogJ: one last nit. Can you add some tests to show how this pipe behaves when called with:

- 'humanize'
- 'device_hours'
- null
- a invalid string

thanks! @bauzer714  this is great work

bauzer714:
@AnalogJ  - updated.  I added tests for "configuration" of each of the 4 requested.

Additionally I added null input values (since null is a number in TS) for the configuration "device_hours" and "humanize". 

image

What's your thoughts on "null hours" vs "0 hours" as the expected output of the "null" input tests? A few options I thought of:

- Leave both as is
- Update the device_hours pipe to return '0 hours'
- Update all configurations of the pipe to return ''
- ???

image

AnalogJ:
if no data is present, maybe "No Data" or "Unknown"

bauzer714:
@AnalogJ  - Updated.   Let me know any additional changes.

image

AnalogJ:
LGTM. Thanks for the PR @bauzer714 ! 🥇

PR #686, opened 2024-09-12
feat: create allow-list for filtering down devices to only a subset

nicjohnson145:
I'm trying to run Scrutiny in Kubernetes. My storage solution involves Longhorn. This means that not all drives are "real" so to speak, some of them are mounted due the fact that a certain pod is present on a host. These pods can move about, which means that disks can come and go.

image

In the above screen shotshot, only sda/sdb/sdc are real physical disks that are plugged into the system, sdd/sde are the result of pods. I would list scrutiny to only bother monitoring the "real" disks, thus an allow list to specify which ones. I cannot use the existing `devices` keyword, because as pods move around I could get more disks than what I have excluded, it would be an endless game of whack-a-mole to exclude the fake disks.

Thus, PR.

### DIFF
--- a/collector/pkg/config/config.go
+++ b/collector/pkg/config/config.go
@@ -20,7 +20,7 @@ import (
 type configuration struct {
 	*viper.Viper
 
-	deviceOverrides []models.ScanOverride
+	deviceOverrides    []models.ScanOverride
 }
 
 //Viper uses the following precedence order. Each item takes precedence over the item below it:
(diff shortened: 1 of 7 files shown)


nicjohnson145:
Also if you want to see a full functional test, I built a new test image off my fork and it works in the real world, not just in unit tests
image

AnalogJ:
This is great, thanks for the PR! 🎉

Very clean implementation

Issue #706, opened 2024-10-31
[FEAT] Allow a delay or different schedules between multiple disks checks


pabsi:
**Is your feature request related to a problem? Please describe.**
The particular issue arises when running a smart check over multiple disks which are connected USB-to-SATA. In my specific case, I have the Quad SATA Hat for the Pi 4, meaning 4 sata disks are connected via 2 USB 3.0 ports. Sometimes when running the smart checks against all 4 drives at once, the USB connection gets reset, and this, in my case, makes the mdadm RAID array fail and mark the devices as failed, and thus removing them from the array. Not a real issue, since I can `--re-add` them later. But it's very inconvenient. Moreover if the smart checks are run daily. See example of `dmesg` logs:

```
[Wed Oct 30 04:00:05 2024] usb 2-2: reset SuperSpeed USB device number 2 using xhci_hcd
[Wed Oct 30 04:00:06 2024] usb 2-1: reset SuperSpeed USB device number 3 using xhci_hcd
[Wed Oct 30 04:00:06 2024] usb 2-1: reset SuperSpeed USB device number 3 using xhci_hcd
[Wed Oct 30 04:00:06 2024] usb 2-1: reset SuperSpeed USB device number 3 using xhci_hcd
[Wed Oct 30 04:00:06 2024] usb 2-1: reset SuperSpeed USB device number 3 using xhci_hcd
[Wed Oct 30 04:00:06 2024] usb 2-1: reset SuperSpeed USB device number 3 using xhci_hcd
[Thu Oct 31 04:00:06 2024] usb 2-2: reset SuperSpeed USB device number 2 using xhci_hcd
[Thu Oct 31 04:00:06 2024] sd 0:0:0:1: [sdb] tag#0 UNKNOWN(0x2003) Result: hostbyte=0x07 driverbyte=DRIVER_OK cmd_age=0s
[Thu Oct 31 04:00:06 2024] sd 0:0:0:1: [sdb] tag#0 CDB: opcode=0x85 85 06 2c 00 00 00 00 00 00 00 00 00 00 00 e5 00
[Thu Oct 31 04:00:07 2024] usb 2-2: USB disconnect, device number 2
[Thu Oct 31 04:00:07 2024] md: super_written gets error=-5
[Thu Oct 31 04:00:07 2024] md/raid10:md0: Disk failure on sdb1, disabling device.
                           md/raid10:md0: Operation continuing on 3 devices.
[...]

PR #725, opened 2024-11-26
Issue 706: Fix time unit

pabsi:
Related to https://github.com/AnalogJ/scrutiny/issues/706

When the original PR was created, I specified in the docs and in the example that the unit was seconds. I was wrong. The function takes milliseconds. This PR makes the documentation correct now.

Other option would be to change the docs to say "milliseconds" instead but I think if this feature is used, people will use higher values (seconds, minutes)  and this will be easier to read.

Thanks

PS: the other format changes was my IDE and the auto-formatting tool :pray:

### DIFF
--- a/collector/pkg/collector/metrics.go
+++ b/collector/pkg/collector/metrics.go
@@ -4,6 +4,12 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"net/url"
+	"os"
+	"os/exec"
+	"strings"
+	"time"
+
 	"github.com/analogj/scrutiny/collector/pkg/common/shell"
 	"github.com/analogj/scrutiny/collector/pkg/config"
 	"github.com/analogj/scrutiny/collector/pkg/detect"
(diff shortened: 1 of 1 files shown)


collector/pkg/collector/metrics.go
```diff
@@ -92,7 +93,7 @@ func (mc *MetricsCollector) Run() error {
 			mc.Collect(device.WWN, device.DeviceName, device.DeviceType)
 
 			if mc.config.GetInt("commands.metrics_smartctl_wait") > 0 {
-				time.Sleep(time.Duration(mc.config.GetInt("commands.metrics_smartctl_wait")) * time.Millisecond)
+				time.Sleep(time.Duration(mc.config.GetInt("commands.metrics_smartctl_wait")) * 1000 * time.Millisecond)
```
AnalogJ:
```suggestion
				time.Sleep(time.Duration(mc.config.GetInt("commands.metrics_smartctl_wait")) * time.Second)
```

Why not just do `time.Second` instead?

pabsi: Well...you are 100% right.
Just changed it!

PR #739, opened 2025-01-09
fix: issue 643 - Fahrenheit values in graph were converted twice

RickZaki:
needed to separate formatting temps from converting values

dashboard was using format method to convert and send Fahrenheit values to chart, then passing the same method into chart formatter causing the Fahrenheit value to be passed in as Celsius and converted again


AnalogJ:
fantastic work @RickZaki ! Thanks for the PR 🥳

AnalogJ (approved):

Issue #678, opened 2024-08-29
[FEAT] Add docker tags for major/minor/patch versions


Quadrubo:
### Description

As of the point of writing this, `v0.8.1` is the latest Scrutiny version.
As I'm manually updating my images and not using something like "latest", I have to change the image tag every new release. It would be nice if I wouldn't have to do that for patch releases (minor releases too).

Currently these tags are provided:
- `v0.8.1` - Version 0.8.1
- `v0.8.0` - Version 0.8.0
- `v0.7.3` - Version 0.7.3
- `latest` - Version 0.8.1

### Desired Behavior

I would propose a tag system like this:
- `v0.8.1` - Version 0.8.1
- `v0.8.0` - Version 0.8.0
- `v0.8`    - Version 0.8.1
- `v0`       - Version 0.8.1
- `latest` - Version 0.8.1

This would allow pinning to specific version and allows the user to choose which semver types they are okay with auto updating.
I for example would have pinned to `v0` and would have automatically gotten the update up to `0.8.1`.
However I won't get the update to a new major version like "v1.x" which would be desired for me.

AnalogJ:
I think the Github workflow docker build step  supports something like that, but I honestly haven't had time to look into it. 

https://github.com/AnalogJ/scrutiny/blob/master/.github/workflows/docker-build.yaml#L43-L50

kaysond:
Closed by #771

PR #771, opened 2025-04-24
Add docker semver tags

RoboMagus:
Implements #678

### DIFF
--- a/.github/workflows/docker-build.yaml
+++ b/.github/workflows/docker-build.yaml
@@ -46,7 +46,9 @@ jobs:
             latest=false
           tags: |
             type=ref,enable=true,event=branch,suffix=-collector
-            type=ref,enable=true,event=tag,suffix=-collector
+            type=semver,pattern=v{{major}}.{{minor}}.{{patch}},suffix=-collector
+            type=semver,pattern=v{{major}}.{{minor}},suffix=-collector
+            type=semver,pattern=v{{major}},suffix=-collector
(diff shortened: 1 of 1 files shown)


AnalogJ:
Hey @RoboMagus this is great, but I'm not certain I understand how the docker image action gets the tag data

RoboMagus:
The magic is provided by [`docker/metadata-action`](https://github.com/docker/metadata-action?tab=readme-ov-file#semver) which for the `semver` type only triggers when a new tag is pushed which this workflow file already captures as a trigger: 
```
on:
  push:
    branches: [ master, beta ]
    # Publish semver tags as releases.
tags: [ 'v*.*.*' ]
```
When e.g. a new version is released as `v0.8.2`, the changes in this PR will add `v0-omnibus` and `v0.8-omnibus` tags in addition to the `v0.8.2-omnibus` tag that was already used. Similarly for the images with `-collector` and `-web` suffixes.

These tags are then picked up by the `docker/build-push-action` action, which already took the `tags` input provided by the `docker/metadata-action` step before this change.

For a working example of the results you could take a look at a very similar [PR that I've done for Zigbee2MQTT](https://github.com/Koenkk/zigbee2mqtt/pull/25530). There's been a couple of releases since that PR got merged. If you take a look at the [packages](https://github.com/Koenkk/zigbee2mqtt/pkgs/container/zigbee2mqtt) you'll see how some images have multiple tags.

AnalogJ:
awesome! I'll go ahead and merge this

PR #773, opened 2025-05-06
Unify docker restart policy among docs and example files

Impact123:
This makes sure the services start up again after a reboot which is kind of important.
I also removed certain flags like `--rm` or `-it` where I thought they were misplaced.


AnalogJ:
thanks for the PR! 🥳

PR #784, opened 2025-07-06
Fix web ui latency

Peppercorn27:
fix web ui latency in situations where cron shedule has been reduced resulting in more data being present in influxDB than expected


Peppercorn27:
@AnalogJ FYI

AnalogJ:
wouldn't this force daily aggregation for shorter timeframes? eg weekly graphs will only have 7 data points?

Screenshot 2025-10-19 at 8 41 05 AM

Peppercorn27:
No idea, I'm not a dev. I'm only interested in the drive health monitoring aspects/alerting + temp. Don't usually look at the frontend.

Mine looks like this after having this modification in place for a few months.

image

AnalogJ:
ah I was completely mistaken. The temp graph wasn't the primary issue here, it was datapoints for the SMART attributes.
Yes, your fix will solve that

PR #805, opened 2025-09-19
Limit Zulip subject length and add URL topic override

merlinz01:
Subjects over 60 characters long, such as the test notification, are rejected by shoutrrr. This truncates the subject to the max length.

Users may want all Scrutiny notifications to be sent to a particular topic rather than whatever Scrutiny happens to decide. This adds the ability to specify a `force_topic` URL parameter to specify that.

### DIFF
--- a/webapp/backend/pkg/notify/notify.go
+++ b/webapp/backend/pkg/notify/notify.go
@@ -428,8 +428,8 @@ func (n *Notify) GenShoutrrrNotificationParams(shoutrrrUrl string) (string, *sho
 			subject = subject[:60]
 		}
 		urlTopic := serviceURL.Query()["force_topic"]
-		if urlTopic != "" {
-			subject = urlTopic
+		if len(urlTopic) > 0 && urlTopic[len(urlTopic)-1] != "" {
+			subject = urlTopic[len(urlTopic)-1]
 		}
 		(*params)["topic"] = subject
 	}
--- a/webapp/backend/pkg/notify/notify.go
+++ b/webapp/backend/pkg/notify/notify.go
@@ -424,6 +424,13 @@ func (n *Notify) GenShoutrrrNotificationParams(shoutrrrUrl string) (string, *sho
 	case "telegram":
 		(*params)["title"] = subject
 	case "zulip":
+		if len(subject) > 60 {
+			subject = subject[:60]
+		}
+		urlTopic := serviceURL.Query()["force_topic"]
+		if urlTopic != "" {
+			subject = urlTopic
+		}
 		(*params)["topic"] = subject
 	}
 
--- a/webapp/backend/pkg/notify/notify.go
+++ b/webapp/backend/pkg/notify/notify.go
@@ -424,7 +424,7 @@ func (n *Notify) GenShoutrrrNotificationParams(shoutrrrUrl string) (string, *sho
 	case "telegram":
 		(*params)["title"] = subject
 	case "zulip":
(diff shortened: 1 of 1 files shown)


AnalogJ (approved):

kaysond:
@merlinz01 can you please check if this is still needed with the new shoutrrr package? If so please rebase and I will review

merlinz01:
I believe it is still needed.

https://github.com/nicholas-fedor/shoutrrr/blob/b8509ad23ca0394fa887b8c0b579aa0da80d0782/pkg/services/chat/zulip/zulip.go#L55-L58

webapp/backend/pkg/notify/notify.go
```diff
@@ -424,6 +424,13 @@ func (n *Notify) GenShoutrrrNotificationParams(shoutrrrUrl string) (string, *sho
 	case "telegram":
 		(*params)["title"] = subject
 	case "zulip":
+		if len(subject) > 60 {
+			subject = subject[:60]
```
kaysond: Please add a logger warning for the subject being truncated.

webapp/backend/pkg/notify/notify.go
```diff
@@ -424,6 +424,13 @@ func (n *Notify) GenShoutrrrNotificationParams(shoutrrrUrl string) (string, *sho
...
+		}
+		urlTopic := serviceURL.Query()["force_topic"]
+		if len(urlTopic) > 0 && urlTopic[len(urlTopic)-1] != "" {
+			subject = urlTopic[len(urlTopic)-1]
+		}
```
kaysond: Why do you want to override the notification subject? This seems like it should be a separate feature request.

kaysond (changes requested):

webapp/backend/pkg/notify/notify.go
merlinz01: It's also related to Zulip subjects but it can be a separate PR if you wish it so.

webapp/backend/pkg/notify/notify.go
merlinz01: Added a warning, as well as checking the rune length rather than the byte length to match how shoutrrr does it.

webapp/backend/pkg/notify/notify.go
kaysond: I'm more asking why it's a problem for Zulip. We could make it a global setting. I take it the problem is Zulip organizes something by "topic" and scrutiny's usage of the message subject as the topic generates a lot of topics?

Couple of thoughts on this:
1. Override should happen before the truncation because it could be longer than 60 chars.
2. Let's call the param "topic"

merlinz01: > I take it the problem is Zulip organizes something by "topic" and scrutiny's usage of the message subject as the topic generates a lot of topics?

Exactly.

> We could make it a global setting.

IMHO it should be configured per notification target so that it can be different for different targets.

merlinz01: > 1. Override should happen before the truncation because it could be longer than 60 chars.
> 2. Let's call the param "topic".

Done.

kaysond:
@merlinz01 sorry one last thing - can you please add the topic param to the example scrutiny yaml file?

merlinz01:
Looks like the `topic` param is already there in the example but would have been ignored previously.

kaysond (approved):

kaysond:
Thanks for your contribution!

Issue #824, opened 2025-11-08
[BUG] `ata_sct_temperature_history` leads to duplicated / missing datapoints


CircuitCoder:
**Describe the bug**
Some disks (e.g. ST3000VX000-1CU166) provides `ata_sct_temperature_history`, but we don't know how the data intervals are aligned against the wall clock. `ST3000VX000-1CU166` gives `"logging_interval_minutes": 94`, which means that if the collector are run per 5 minutes:
- Every 5 minutes, a new data point is created for each of the recent data points, because `currTime - 94 * idx` ticks forward by 5 minutes.
- Worse, because 5 is coprime with 94, each data point will eventually create 94 different points, filling the entire interval.

Also, because we almost never land on the logging interval boundary, the most recent value is almost always `null`. This means that we'll almost never get the most recent temperature for this disk (because we ignore current temperature if `ata_sct_temperature_history` is present).

**Expected behavior**
I'm not sure what's the better way to solve this. Most obviously we might want to log current temperature in spite of the existence of `ata_sct_temperature_history`.

Also, we might want to only write data from `ata_sct_temperature_history` if no data point exists within the logging interval calculated with current wall time (i.e. `[currTime - loggingInterval * (idx + 1), currTime - loggingInterval * idx)`). Because the true interval boundary will always lies inside this interval, this means that each data point will at most be written once.

Alternatively, manually align each data point to loggingInterval. Thus no matter the wall clock time, each data point will always land on the same time. However if the disk somehow decide to change its reported `loggingInterval` then this breaks. I haven't seen any disk doing this though.

## Versions

```
smartctl 7.5 2025-04-30 r5714 [x86_64-linux-6.15.2-arch1-1] (local build)
scrutiny built from source, 7c35d59552a19b167f54aed081f8b7e763fe46c8
```

PR #825, opened 2025-11-08
Fixes duplicated / missing recent temperature datapoints caused by `ata_sct_temperature_history`

CircuitCoder:
This commit makes two modification to smart temperature logging:
- Even if ata_sct_temperature_history is present, log current time.
- Forcefully align each ata_sct_temperature_history data point according to logging interval reported in SMART. This prevents duplicated datapoints.

This is one of the solution mentioned in #824.

Also, this PR / issue should be independent from #494 / #557. That issue seems to stem from unreliable `ata_sct_temperature_history`, and the solution proposed in the PR was to add a switch to turn off `ata_sct_temperature_history` sampling. This PR tries to fix some logging issue when (semi-)reliable `ata_sct_temperature_history` is reported without completely turning it off.

### DIFF
--- a/webapp/backend/pkg/database/scrutiny_repository_temperature.go
+++ b/webapp/backend/pkg/database/scrutiny_repository_temperature.go
@@ -22,9 +22,11 @@ func (sr *scrutinyRepository) SaveSmartTemperature(ctx context.Context, wwn stri
 				continue
 			}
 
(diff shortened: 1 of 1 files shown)


kaysond:
@CircuitCoder - can you please rebase this?

CircuitCoder:
> @CircuitCoder - can you please rebase this?

@kaysond Sure, now rebased to the current master (761014a93fb8e072103962cfc119bfd4d059f5be)

kaysond:
@CircuitCoder thanks for your contribution!

PR #845, opened 2026-01-22
fix(smart): prevent false failures from corrupted ATA device statistics

### LINKED ISSUE #84: Debian/Ubuntu/Centos Install packages using FPM
Heya, I made install scripts for scrutiny for a seedbox toolset I'm contributing to. There's also a removal script. I've tested this on a VM and it works insofar as I can't get actual smart data to show because VM is too virtualised.

You can find the files here https://github.com/liaralabs/swizzin/pull/418/files

They should be extremely easy to adapt (i.e. remove `$log` redirects or define the variable yourself and use a normal `apt` function, that should be enough I think), probably also for non-Debian derivatives. You really just need curl and a package manager.

I also do away with crontab as it's a pain and use systemd timers instead

Let me know if you'd like me to adapt an installer for your specific application that's not dependent on a toolset!

Starosdev:
## Summary

Fixes false failure alerts caused by impossibly high ATA Device Statistics values (e.g., 420 billion mechanical failures reported in issue #84).

- Add `AttributeStatusInvalidValue` constant (value 8) for corrupted data detection
- Add `MaxReasonableFailureCount` threshold (1 million) - values above this are marked invalid, not failed
- Add device status propagation to `ProcessAtaDeviceStatistics` (was missing, unlike other processors)
- Fix `notify.go` to handle string-based devstat attribute IDs when filtering by critical attributes
- Add config-based ignore lists (`failures.ignored.devstat`) for user-configurable attribute suppression
- Add comprehensive unit tests for all new functionality

## Test plan

- [x] All 35 measurement tests pass
- [x] All 16 notify tests pass
- [x] Build verification passes
- [ ] Manual testing with drives reporting high devstat values

Closes #84

Generated with [Claude Code](https://claude.ai/code)


Starosdev:
Closing - accidentally created against upstream instead of fork (Starosdev/scrutiny)

PR #847, opened 2026-01-26
Human-readable values for data written/read attributes

jwidess:
This small PR adds a human-readable data size in brackets next to the SMART attributes related to data written/read. This makes it a bit easier to understand the amount of data written/read on a drive and mirrors the format `smartctl` uses in their reporting as well.

**Before:**
image

**After:**
image

---

### DIFF
--- a/webapp/frontend/src/app/data/mock/device/details/sdb.ts
+++ b/webapp/frontend/src/app/data/mock/device/details/sdb.ts
@@ -30,6 +30,17 @@ export const sdb = {
             'power_on_hours': 1730,
             'power_cycle_count': 9,
             'attrs': {
+                '241': {
+                    'attribute_id': 241,
+                    'value': 100,
+                    'thresh': 0,
+                    'worst': 100,
+                    'raw_value': 1953125000,
+                    'raw_string': '1953125000',
+                    'when_failed': '',
+                    'transformed_value': 0,
+                    'status': 0
+                },
                 '1': {
                     'attribute_id': 1,
                     'value': 100,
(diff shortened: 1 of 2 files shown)


kaysond:
@jwidess -

1. Can you please disclose your use of AI on this, per our [AI policy](https://github.com/AnalogJ/scrutiny/blob/master/AI_POLICY.md)
2. It looks like you've hardcoded 512 here but I'm fairly certain the smart attributes (for ATA at least) are in blocks. So it could be 4096 for some drives
3. Please rebase this

jwidess:
@kaysond 1. Absolutely, I used VS Code Copilot with a Gemini 3 Pro agent for the majority of these changes and to identify where exactly it needed to be implemented. I felt that for the simplicity of this commit, this was easy and acceptable. And I did test this on two separate machines to verify that my changes functioned and displayed correctly.
2. Yes, 512 is hardcoded, that was just the math I copied over from figuring out the conversion on my NVMe drive without thinking about different sector block sizes! I will take a look at addressing this to dynamically use the correct block size dependent on the drive, because otherwise, you are right, this will report a very wrong value for 4096.
Feel free to either close this PR and I will make another once I've implemented and tested that, or give me a few days/weeks to have some time to review that. 
3. Done.

kaysond:
> @kaysond 1. Absolutely, I used VS Code Copilot with a Gemini 3 Pro agent for the majority of these changes and to identify where exactly it needed to be implemented. I felt that for the simplicity of this commit, this was easy and acceptable. And I did test this on two separate machines to verify it functioned and displayed correctly. 2. Yes, 512 is hardcoded, that was just the math I copied over from figuring out the conversion on my NVMe drive without thinking about different sector block sizes! I will take a look at addressing this to dynamically use the correct block size dependent on the drive, because otherwise, you are right, this will report a very wrong value for 4096. Feel free to either close this PR and I will make another once I've implemented and tested that, or give me a few days/weeks to have some time to review that. 3. Done.

Thanks. We can leave this one open. Feel free to ping me when you're ready for review

PR #861, opened 2026-02-05
Add devcontainer

### LINKED ISSUE #853: [FEAT] Dev containers
Setting up the full development stack isn't straight forward for new contributors. I suggest adding Dev Container support (`.devcontainer`). This would provide a standardized, "ready-to-code" environment, ensuring everyone has the same tools and versions pre-configured.

mcarbonne:
Solve #853.
Add a minimal working devcontainer environment.

Note: This PR is based on an initial draft generated by Gemini 3 Pro.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,28 @@
+{
+  "name": "Scrutiny Dev Env",
+  "dockerComposeFile": "docker-compose.yml",
+  "service": "app",
+  "workspaceFolder": "/workspaces/scrutiny",
+
+  "features": {
+    "ghcr.io/devcontainers/features/go:1": "1.25",
+    "ghcr.io/devcontainers/features/node:1": "lts"
+  },
+
+  "onCreateCommand": "sudo apt-get update && sudo apt-get install -y smartmontools iputils-ping",
+
+  "customizations": {
+    "vscode": {
+      "extensions": [
+        "golang.go",
+        "dbaeumer.vscode-eslint",
+        "esbenp.prettier-vscode"
+      ]
+    }
+  },
+
+  "forwardPorts": [8080, 8086],
+
+  "postCreateCommand": "bash .devcontainer/setup.sh",
+  "remoteUser": "vscode"
+}
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -24,5 +24,7 @@
   "forwardPorts": [8080, 8086],
 
   "postCreateCommand": "bash .devcontainer/setup.sh",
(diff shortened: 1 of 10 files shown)


kaysond:
@mcarbonne unfortunately I can't seem to get this to work, but I suspect its because my home directory is mounted on nfs. Seems there are some workarounds (https://github.com/microsoft/vscode-remote-release/issues/9007#issuecomment-1746626507) but nothing that would really make sense to add to a project config. Let me know if you have any ideas..

Only other request would be could you add a `.vscode/extensions.json` to recommend the golang and devcontainer extensions? (the former is needed for the debug config)

kaysond:
Some more info about the devcontainer setup here: https://stackoverflow.com/questions/68937464/running-devcontainer-as-local-user-instead-of-root

Adding `"updateRemoteUserUID": false`, and setting `remoteUser` and `containerUser` to `root` helped me get the container running. Can you try that and let me know if it breaks anything for you?

mcarbonne:
> Some more info about the devcontainer setup here: https://stackoverflow.com/questions/68937464/running-devcontainer-as-local-user-instead-of-root
> 
> Adding `"updateRemoteUserUID": false`, and setting `remoteUser` and `containerUser` to `root` helped me get the container running. Can you try that and let me know if it breaks anything for you?

Sorry for the late reply, I wasn't available.
I’ve rebased the PR and updated CONTRIBUTING.md.

Regarding the workaround suggestion, I'm not sure it's a good idea. It messes up permissions on the host system (new files are created as root instead of the correct user).

I saw that there is a "Dev Containers: Clone Repository in Container Volume" option for NFS issues, among other things. Could you try it and let me know if it works for you?

kaysond:
> Regarding the workaround suggestion, I'm not sure it's a good idea. It messes up permissions on the host system (new files are created as root instead of the correct user).

If you're run into the docker daemon as root, yes. But if you're running rootless, then root in the container maps back to your own uid on the host which is what you want. 

I did a bit of research - I don't think we can just have one devcontainer.json. Different settings are needed for different setups. Fortunately, you can have multiple directories in .devcontainer, each having their own devcontainer.json file, and users can pick between them from the vscode menu when they select reopen in container. See https://stackoverflow.com/questions/58617041/multiple-devcontainers-in-vscode-remote

I'd propose 3 configs - docker (what you have now), docker.rootless (what I used), and podman, which needs some special config: https://medium.com/@guillem.riera/making-visual-studio-code-devcontainer-work-properly-on-rootless-podman-8d9ddc368b30

kaysond (approved):

kaysond:
@mcarbonne - thanks for the contribution!

PR #869, opened 2026-02-06
fix: page smart attribute queries with limit and sort

AlliotTech:
- use limit instead of tail for paging
- sort by _time desc before paging to ensure latest entries


kaysond (approved):

kaysond:
Thanks for the contribution!

PR #875, opened 2026-02-06
Update go to 1.25

### LINKED ISSUE #872: Bump go to 1.25.0
Fixes #584. See #585.

Need to update in many places, including, but not limited to:

https://github.com/AnalogJ/scrutiny/blob/0565962a1445eea0a7aa23f87fd83686a843fcdd/.github/workflows/ci.yaml#L105

https://github.com/AnalogJ/scrutiny/blob/0565962a1445eea0a7aa23f87fd83686a843fcdd/docker/Dockerfile#L15

https://github.com/AnalogJ/scrutiny/blob/0565962a1445eea0a7aa23f87fd83686a843fcdd/docker/Dockerfile.collector#L7

https://github.com/AnalogJ/scrutiny/blob/0565962a1445eea0a7aa23f87fd83686a843fcdd/docker/Dockerfile.web#L14

https://github.com/AnalogJ/scrutiny/blob/0565962a1445eea0a7aa23f87fd83686a843fcdd/CONTRIBUTING.md?plain=1#L14

https://github.com/AnalogJ/scrutiny/blob/0565962a1445eea0a7aa23f87fd83686a843fcdd/.github/workflows/release.yaml#L99

See https://github.com/search?q=repo%3AAnalogJ%2Fscrutiny%201.20&type=code

mcarbonne:
Fix #872


mcarbonne:
@kaysond this PR is ready for review.
This PR is split into distinct commits to facilitate the review.
I also added a go linter (non-blocking for now to avoid changing too much code).
Later, this should be configured to fail on error by removing those lines:
```
        with:
          args: --issues-exit-code=0
```

.github/workflows/ci.yaml
```diff
@@ -21,7 +21,6 @@ jobs:
   test-backend:
     name: Test Backend
     runs-on: ubuntu-latest
-    container: ghcr.io/packagrio/packagr:latest-golang
```
kaysond, line 24: Why did this need to be removed? Not sure, to be honest, why it was being used.

.github/workflows/golang-lint.yml
```diff
@@ -0,0 +1,21 @@
...
+
+jobs:
+  golangci:
+    name: lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+      - uses: actions/setup-go@v6
+        with:
+          go-version: stable
```
kaysond: should this not match 1.25 as elsewhere?

collector/pkg/collector/base.go
```diff
@@ -14,17 +15,6 @@ type BaseCollector struct {
 	logger *logrus.Entry
 }
 
-func (c *BaseCollector) getJson(url string, target interface{}) error {
```
kaysond, line 17: I'm guessing this isn't used anywhere? I'm inclined to leave it in regardless

.golangci.yml
kaysond, line 1: Will this lint formatting issues? it's not clear to me from the golangci-lint docs, which seem to allow you to enable `gofmt` explicitly, but also say the default "uses standard Go formatting".

kaysond (commented):
@mcarbonne - this is a fantastic PR. Thank you for the contribution and for making it so easy to review!

Really like the addition of the linting check. I know I still need to review the devcontainer PR. This might go without saying, but we should make sure to align the settings such that the container, vscode, etc all follow the same rules.

Finally - It may take a bit to merge this one. I want to update the workflows and docs first to point people to `latest` tags, which I'll change to be the most recent release. Currently everything suggests `master` which gets pushed on every commit.

.golangci.yml
mcarbonne: Regarding formatting, I have just enabled gofmt and goimports in [this commit](https://github.com/mcarbonne/scrutiny/commit/c8e2d5eeb46e841246347fdff371b713b0447db4).

I'm using vscode which automatically formats files on save.

collector/pkg/collector/base.go
mcarbonne: Indeed, the linter was complaining about unused code. I can re-add it, but once the linter is set to mandatory, this will need to be addressed.

.github/workflows/golang-lint.yml
mcarbonne: Updated in [this commit](https://github.com/mcarbonne/scrutiny/commit/c8e2d5eeb46e841246347fdff371b713b0447db4).

.golangci.yml
mcarbonne: Regarding formatting, I have just enabled gofmt and goimports in [this commit](https://github.com/mcarbonne/scrutiny/commit/c8e2d5eeb46e841246347fdff371b713b0447db4).

I'm using vscode which automatically formats files on save.

.github/workflows/ci.yaml
mcarbonne: It seems that `ghcr.io/packagrio/packagr:latest-golang` is outdated and is only related to releases (and not tests). I don't know the impact on release process but building docker images is still working fine (I'm using go 1.25 in my fork). At some point, maybe we should replace packagr with a more standard release process ?
Without removing this line, I wasn't able to fix the tests.

kaysond:
Thanks again!

Issue #871, opened 2026-02-06
Build a specific version of smartmontools instead of using the debian package


kaysond:
Address issues like #836 #522 by building smartmontools ourselves in the collector and omnibus containers. This will allow us to use more modern versions of the drivedb vs the OS package.

v7.5 is most recent as of now, and the JSON format changed at some point, making this a breaking change. It probably makes sense to include this in a 1.0.0.

Xyz00777:
i think its a good idea to bring this into a major version update based on the breaking change

queeup:
I am using `post_start` to update `drivedb`.

```yaml
services:
  scrutiny:
    container_name: scrutiny
    image: ghcr.io/analogj/scrutiny:master-omnibus
    post_start:
      - command: [ "/usr/sbin/update-smart-drivedb", "--no-verify" ]
    cap_add:
      - SYS_RAWIO
...
```

FurkanVG:
Hi all,
to not be confused, these Command timeout read issues still exist and there is no current build that fixes this?

kaysond:
> Hi all, to not be confused, these Command timeout read issues still exist and there is no current build that fixes this?

Its a smartmontools issue, not scrutiny. Should be fixable with a collector config file. See #885

Semmu:
> by building smartmontools ourselves in the collector and omnibus containers

Hi, just curious, what is the plan here exactly? A multi-stage docker build which clones the smartmontools repo and builds it in a builder stage? Wouldn't simply downloading a pinned release from [here](https://github.com/smartmontools/smartmontools-builds/releases) also be sufficient and simpler to implement? (With the same unfortunate risk of relying on an external dependency.)

kaysond:
> > by building smartmontools ourselves in the collector and omnibus containers
> 
> Hi, just curious, what is the plan here exactly? A multi-stage docker build which clones the smartmontools repo and builds it in a builder stage?

Essentially, yes.

> Wouldn't simply downloading a pinned release from [here](https://github.com/smartmontools/smartmontools-builds/releases) also be sufficient and simpler to implement? (With the same unfortunate risk of relying on an external dependency.)

Those builds are all pre-release and they don't have anything for ARM unfortunately

Semmu:
> Those builds are all pre-release and they don't have anything for ARM unfortunately

Ohh okay, didn't realize this limitation, the plan makes sense now.

mill1000:
[...]

Issue #232, opened 2022-05-08
[FEAT] Distroless base images


AnalogJ:
Migrate to a distroless style base image. No need for addl software in the collector images.

AnalogJ:
https://hub.docker.com/hardened-images/catalog might be an interesting alternative as well

PR #924, opened 2026-02-14
Build and distribute latest smartmontools 7.5 as part of image

slydetector:
Fix for issue https://github.com/AnalogJ/scrutiny/issues/871

AI Disclosure:  Google Antigravity with Claude Opus 4.6 was used to help author this PR.

Testing done:
- Created both collector and omnibus images and verified `smartctl` is version 7.5 and the drive database is the latest.

Unrelated changes:
- Fixed docker warnings about inconsistent case of `FROM ... AS ...`

No automated build/tests ran on my fork when creating a PR so I'm hoping something automated will run and surface any additional issues with this PR against `AnalogJ/scritiny`.

### DIFF
--- a/docker/Dockerfile.smartmontools
+++ b/docker/Dockerfile.smartmontools
@@ -1,0 +1,20 @@
+########################################################################################################################
+# Smartmontools Builder
+# Builds smartctl from source as a static binary.
+# Updates the drive database to include the latest drive models since it can change between releases.
+# Used as a shared build stage by Dockerfile and Dockerfile.collector.
+########################################################################################################################
+FROM debian:bookworm-slim
+ARG SMARTMONTOOLS_VER=7.5
+RUN apt-get update && DEBIAN_FRONTEND=noninteractive \
+    apt-get install -y --no-install-recommends \
+    ca-certificates curl gcc g++ gnupg make \
+    && rm -rf /var/lib/apt/lists/*
+RUN curl -L "https://github.com/smartmontools/smartmontools/releases/download/RELEASE_$(echo ${SMARTMONTOOLS_VER} | tr '.' '_')/smartmontools-${SMARTMONTOOLS_VER}.tar.gz" -o /tmp/smartmontools.tar.gz \
+    && tar -xzf /tmp/smartmontools.tar.gz -C /tmp \
+    && cd /tmp/smartmontools-${SMARTMONTOOLS_VER} \
+    && ./configure --prefix=/usr LDFLAGS='-static' --without-libcap-ng --without-libsystemd \
+    && make -j"$(nproc)" \
+    && make install \
+    && /usr/sbin/update-smart-drivedb \
+    && rm -rf /tmp/smartmontools*
--- a/docker/Dockerfile.smartmontools
+++ b/docker/Dockerfile.smartmontools
@@ -4,7 +4,7 @@
 # Updates the drive database to include the latest drive models since it can change between releases.
 # Used as a shared build stage by Dockerfile and Dockerfile.collector.
 ########################################################################################################################
-FROM debian:bookworm-slim
+FROM debian:trixie-slim
 ARG SMARTMONTOOLS_VER=7.5
 RUN apt-get update && DEBIAN_FRONTEND=noninteractive \
     apt-get install -y --no-install-recommends \
--- a/docker/Dockerfile.smartmontools
+++ b/docker/Dockerfile.smartmontools
@@ -3,8 +3,6 @@
 # - Builds smartctl from source as a static binary.
 # - Updates the drive database to include the latest drive models since it can change between releases.
 # - Used as a shared build stage by Dockerfile and Dockerfile.collector.
-# - This Dockerfile has to reside in a directory by itself since build-contexts in the various github workflow files
-#   expect a directory as a build context.
 ########################################################################################################################
 FROM debian:trixie-slim
 ARG SMARTMONTOOLS_VER=7.5
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -151,6 +151,8 @@ jobs:
           context: .
           file: docker/Dockerfile
           push: false
+          build-contexts: |
+            smartmontools-build=docker/Dockerfile.smartmontools
           cache-from: type=gha
           cache-to: type=gha,mode=max
       - name: Build collector
@@ -160,6 +162,8 @@ jobs:
           context: .
           file: docker/Dockerfile.collector
           push: false
+          build-contexts: |
+            smartmontools-build=docker/Dockerfile.smartmontools
           cache-from: type=gha
           cache-to: type=gha,mode=max
       - name: Build web
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -134,6 +134,11 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: read
+    services:
+      registry:
+        image: registry:2
+        ports:
+          - 5000:5000
 
     steps:
       - name: Checkout repository
@@ -144,6 +149,20 @@ jobs:
           platforms: 'arm64,arm'
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3
+        with:
+          driver-opts: network=host
+          buildkitd-config-inline: |
+            [registry."localhost:5000"]
+              http = true
+              insecure = true
+      - name: Build smartmontools
+        uses: docker/build-push-action@v6
+        with:
+          context: .
+          file: docker/Dockerfile.smartmontools
+          platforms: linux/amd64,linux/arm64,linux/arm/v7
+          push: true
+          tags: localhost:5000/smartmontools-build:latest
       - name: Build omnibus
         uses: docker/build-push-action@v6
         with:
@@ -152,7 +171,7 @@ jobs:
           file: docker/Dockerfile
           push: false
           build-contexts: |
-            smartmontools-build=docker/smartmontools
+            smartmontools-build=docker-image://localhost:5000/smartmontools-build:latest
           cache-from: type=gha
           cache-to: type=gha,mode=max
       - name: Build collector
@@ -163,7 +182,7 @@ jobs:
           file: docker/Dockerfile.collector
           push: false
           build-contexts: |
-            smartmontools-build=docker/smartmontools
+            smartmontools-build=docker-image://localhost:5000/smartmontools-build:latest
           cache-from: type=gha
           cache-to: type=gha,mode=max
       - name: Build web
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -152,7 +152,7 @@ jobs:
           file: docker/Dockerfile
           push: false
           build-contexts: |
-            smartmontools-build=docker/Dockerfile.smartmontools
+            smartmontools-build=docker/smartmontools
           cache-from: type=gha
           cache-to: type=gha,mode=max
       - name: Build collector
@@ -163,7 +163,7 @@ jobs:
           file: docker/Dockerfile.collector
           push: false
           build-contexts: |
-            smartmontools-build=docker/Dockerfile.smartmontools
+            smartmontools-build=docker/smartmontools
           cache-from: type=gha
           cache-to: type=gha,mode=max
       - name: Build web
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -134,11 +134,6 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: read
-    services:
-      registry:
-        image: registry:2
-        ports:
-          - 5000:5000
 
     steps:
       - name: Checkout repository
@@ -149,20 +144,6 @@ jobs:
           platforms: 'arm64,arm'
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3
-        with:
-          driver-opts: network=host
-          buildkitd-config-inline: |
-            [registry."localhost:5000"]
-              http = true
-              insecure = true
-      - name: Build smartmontools
-        uses: docker/build-push-action@v6
-        with:
-          context: .
-          file: docker/Dockerfile.smartmontools
-          platforms: linux/amd64,linux/arm64,linux/arm/v7
-          push: true
-          tags: localhost:5000/smartmontools-build:latest
       - name: Build omnibus
         uses: docker/build-push-action@v6
         with:
@@ -170,8 +151,6 @@ jobs:
           context: .
           file: docker/Dockerfile
           push: false
-          build-contexts: |
-            smartmontools-build=docker-image://localhost:5000/smartmontools-build:latest
           cache-from: type=gha
           cache-to: type=gha,mode=max
       - name: Build collector
@@ -181,8 +160,6 @@ jobs:
           context: .
           file: docker/Dockerfile.collector
           push: false
-          build-contexts: |
-            smartmontools-build=docker-image://localhost:5000/smartmontools-build:latest
           cache-from: type=gha
           cache-to: type=gha,mode=max
       - name: Build web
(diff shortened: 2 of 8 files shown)


slydetector:
@kaysond Does this need some kind of approval to get the build going?

docker/Dockerfile.smartmontools
```diff
@@ -0,0 +1,20 @@
...
+    apt-get install -y --no-install-recommends \
+    ca-certificates curl gcc g++ gnupg make \
+    && rm -rf /var/lib/apt/lists/*
+RUN curl -L "https://github.com/smartmontools/smartmontools/releases/download/RELEASE_$(echo ${SMARTMONTOOLS_VER} | tr '.' '_')/smartmontools-${SMARTMONTOOLS_VER}.tar.gz" -o /tmp/smartmontools.tar.gz \
+    && tar -xzf /tmp/smartmontools.tar.gz -C /tmp \
+    && cd /tmp/smartmontools-${SMARTMONTOOLS_VER} \
+    && ./configure --prefix=/usr LDFLAGS='-static' --without-libcap-ng --without-libsystemd \
+    && make -j"$(nproc)" \
+    && make install \
+    && /usr/sbin/update-smart-drivedb \
```
kaysond, line 19: We shouldn't blindly update drivedb. It should be versioned just like the tool itself.

kaysond (changes requested):
Thanks for the PR! Did you try running the containers? The json output in smartmontools v7.5 is different so I expect things won't actually work. The CI tests run on the binary not in docker, so they'll still pass.

Another thing we should consider as part of this is changing the base image. Not sure what the smartmontools dependencies are but I think using debian Trixie would be the right choice.

slydetector:
Can you provide something concrete as to why you believe the JSON in smartmontools 7.5 is a breaking change?  I went through the smartmontools changelog from 7.3 to 7.5 and didn't note any breaking changes.

Changing the base image seems orthogonal to just baking in smartmontools. It would be safer to change that across all three images (not just the two that are being changed in this PR) under a separate issue.

I did a spot check by running the image, but let me test it against my prod deployment for a few days to make sure.

docker/Dockerfile.smartmontools
slydetector: AFAICT, drivedb changes independently of the version of smartmontools, that is why it is always updated to the "latest" as part of the build.

IIRC, `update-smart-drivedb` makes sure that the "latest" version within the version that was built is used.

kaysond: > AFAICT, drivedb changes independently of the version of smartmontools, that is why it is always updated to the "latest" as part of the build.
> 
> IIRC, `update-smart-drivedb` makes sure that the "latest" version within the version that was built is used.

Correct, it's independent, but we should still choose the version. Without making it explicit, it's not immediately clear from the Dockerfile which version is installed. For debug purposes it's useful to know.

I'd prefer having to bump an explicit version manually than dig through the container or build logs to know which version it is.

kaysond:
> Can you provide something concrete as to why you believe the JSON in smartmontools 7.5 is a breaking change? I went through the smartmontools changelog from 7.3 to 7.5 and didn't note any breaking changes.

A few things:

* https://github.com/smartmontools/smartmontools/issues/168
* The release notes mention a regression in 7.4
* original maintainer told me things have changed since 7.2

> Changing the base image seems orthogonal to just baking in smartmontools. It would be safer to change that across all three images (not just the two that are being changed in this PR) under a separate issue.

Fair enough. My concern was mainly that having to update the json parsing would be a breaking change, and I'd want to release the base image change together.

> I did a spot check by running the image, but let me test it against my prod deployment for a few days to make sure.

If there's no json change, this becomes way easier! Thanks!

slydetector:
Oops, I did update my prod instance to use the baked in smartmontools images and they have been running without a hitch the last few days.  Testing wise, things look good on my end.

I'm not seeing any changes to the JSON format which are breaking things. https://github.com/smartmontools/smartmontools/issues/168 seems to pre-date 7.3 and the scrutiny codebase seems to reference scsi keys with the `scsi_` prefix correctly.

kaysond:
> Oops, I did update my prod instance to use the baked in smartmontools images and they have been running without a hitch the last few days. Testing wise, things look good on my end.
> 
> I'm not seeing any changes to the JSON format which are breaking things. [smartmontools/smartmontools#168](https://github.com/smartmontools/smartmontools/issues/168) seems to pre-date 7.3 and the scrutiny codebase seems to reference scsi keys with the `scsi_` prefix correctly.

Were you using 7.3 before?

Will need to see if the code is trying to get scsi temperature, which was removed in 7.4. I'm guessing not

docker/Dockerfile.smartmontools
slydetector: I understand the need for a reproducible build, but the latest version of the drivedb within a specific smartmontools version only seems obtainable by the actual git commit SHA on a specific drivedb branch.  For example, the version attached to the latest 7.5 is 6088 (monotonically increasing but jumps forward skipping numbers):

example json snippet from smartctl output:
```
time="2026-02-17T16:30:41-06:00" level=info msg="Executing command: smartctl --info --json /dev/sdb" type=metrics
{
  "json_format_version": [
    1,
    0
  ],
  "smartctl": {
    "version": [
      7,
      5
    ],
    ...
    "drive_database_version": {
      "string": "7.5/6088"
    },
  },
...
```
I don't see a straightforward way to retrieve version 6088 via explicit URL. Using the comit SHA seems undesirable and unwieldy to maintain.  For debugging purposes, the drivedb version is in the log output.

slydetector:
> > Oops, I did update my prod instance to use the baked in smartmontools images and they have been running without a hitch the last few days. Testing wise, things look good on my end.
> > I'm not seeing any changes to the JSON format which are breaking things. [smartmontools/smartmontools#168](https://github.com/smartmontools/smartmontools/issues/168) seems to pre-date 7.3 and the scrutiny codebase seems to reference scsi keys with the `scsi_` prefix correctly.
> 
> Were you using 7.3 before?
> 
> Will need to see if the code is trying to get scsi temperature, which was removed in 7.4. I'm guessing not

Yes, I was using 7.3.

kaysond:
Ok let me give this some thought. The script just grabs the latest commit on the drivedb/release# branch and verifies the pgp signature. We could do that ourselves straight from a commit. A github action could periodically check for updates and submit a PR.

I'll also do a quick sanity check on the json between 7.2 and 7.5.

In the meantime, could you please update the two docker workflow files to build the smartmontools container? They don't use the makefile so a step will have to be added (in a few places). LMK if you have any questions

slydetector:
> Ok let me give this some thought. The script just grabs the latest commit on the drivedb/release# branch and verifies the pgp signature. We could do that ourselves straight from a commit. A github action could periodically check for updates and submit a PR.

Honestly, I'm of the opinion this is a case of YAGNI. As a user, I just want the latest and greatest and I just want it to work when I pull down the the latest stable image. There are plenty of cases where I would insist on a reproducible build, but this just isn't one of them. Feel free to close this PR if you think otherwise.
 
> I'll also do a quick sanity check on the json between 7.2 and 7.5.
> 
> In the meantime, could you please update the two docker workflow files to build the smartmontools container? They don't use the makefile so a step will have to be added (in a few places). LMK if you have any questions

kaysond:
> > Ok let me give this some thought. The script just grabs the latest commit on the drivedb/release# branch and verifies the pgp signature. We could do that ourselves straight from a commit. A github action could periodically check for updates and submit a PR.
> 
> Honestly, I'm of the opinion this is a case of YAGNI. As a user, I just want the latest and greatest and I just want it to work when I pull down the the latest stable image. There are plenty of cases where I would insist on a reproducible build, but this just isn't one of them. Feel free to close this PR if you think otherwise.
> 
> > I'll also do a quick sanity check on the json between 7.2 and 7.5.
> > In the meantime, could you please update the two docker workflow files to build the smartmontools container? They don't use the makefile so a step will have to be added (in a few places). LMK if you have any questions

I'm inclined to agree with you on the drivedb but just want to think it through.

The workflows do need to be updated for me to merge this though.

kaysond:
@slydetector - I'm not seeing any json changes between 7.2 and 7.5 on the drives I have, and without an easy way to version drivedb I agree it's not worth the effort for the sake of reproducible builds.

Can you please rebase this and add a step to build the smartmontools container in https://github.com/AnalogJ/scrutiny/blob/master/.github/workflows/docker-build.yaml and https://github.com/AnalogJ/scrutiny/blob/master/.github/workflows/docker-nightly.yaml

slydetector:
Sure thing. Will circle back in a couple of days.

slydetector:
@kaysond Things should be in good shape to merge now.

kaysond:
@slydetector - just took a look at this. I think you'll have to add an extra build action for the smartmontools container before the omnibus/collector (with `push: false`)

slydetector:
@kaysond Why is the workflow continually waiting on approval from a maintainer? This introduces what seems like unnecessary friction to the build feedback cycle.

kaysond (approved):

kaysond:
> @kaysond Why is the workflow continually waiting on approval from a maintainer? This introduces what seems like unnecessary friction to the build feedback cycle.

I don't have access to the repo settings, unfortunately. It's presumably set up to require approval for first time contributors (possibly for new github accounts only; not sure how new yours is) - https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#controlling-changes-from-forks-to-workflows-in-public-repositories

kaysond:
@slydetector is there a reason you opted to add a registry instead of uploading/downloading the images (or even just the two files) as artifacts? 

See https://docs.docker.com/build/ci/github-actions/share-image-jobs/

slydetector:
> @slydetector is there a reason you opted to add a registry instead of uploading/downloading the images (or even just the two files) as artifacts?
> 
> See https://docs.docker.com/build/ci/github-actions/share-image-jobs/

@kaysond Not my area of expertise, but this is why Claude Opus preferred the registry instead:

> 
> The artifact sharing approach from the Docker docs is designed for passing images between separate jobs on different runners. In this case, the smartmontools and omnibus/collector builds all happen within the same job, so artifacts aren't the issue.
> 
> The problem is that docker/setup-buildx-action creates a builder using the docker-container driver, which runs BuildKit inside an isolated container. This means the buildx builder cannot see images in the host Docker daemon's store. Even if we build smartmontools with a plain docker build -t smartmontools-build . in a prior step, when buildx encounters FROM smartmontools-build during the multi-platform omnibus/collector build, it can't resolve it — the image is on the host daemon, but buildx is running in its own container.
> 
> Additionally, docker load / outputs: type=docker (tarball export) doesn't support multi-platform images, so we can't export and re-import a single tarball covering amd64/arm64/arm.
> 
> The local registry gives the buildx builder a network-reachable endpoint to pull the pre-built multi-platform smartmontools image from, which is the [standard pattern](https://github.com/docker/buildx/discussions/1382) for this kind of setup. The alternative would be to inline the smartmontools build as a FROM ... AS smartmontools-build stage directly in each Dockerfile, but that duplicates the build instructions across files.
> 

Happy to change course here if this is just flat out incorrect or there is a simpler approach.

AnalogJ:
@slydetector If I'm reading this correctly, your intention is to reuse the smartmontools base image in multiple build jobs. The way this is setup won't solve that problem imo, as each job has its own isolated local registry.

What you might want to do instead is add a new job to build the smartmontools base image and upload it to the GHCR (and benefit from caching)
Then have the web/omnibus/collector images depend on that job and pull the GHCR image during their builds 

I think that should work

kaysond:
> @slydetector If I'm reading this correctly, your intention is to reuse the smartmontools base image in multiple build jobs. The way this is setup won't solve that problem imo, as each job has its own isolated local registry.
> 
> What you might want to do instead is add a new job to build the smartmontools base image and upload it to the GHCR (and benefit from caching) Then have the web/omnibus/collector images depend on that job and pull the GHCR image during their builds
> 
> I think that should work

We'd have to be careful not to clobber tags between workflow runs. And if I'm not mistaken, workflows run in a PR wouldn't have the registry credentials so pushing to GCHR would fail wouldn't it?

I think Claude is confusing a couple of separate things: you can't set the buildx output to tar if you are doing a multiplatform build. At least according to the docs. I vaguely recall that it does work now but just gives you a separate directory per platform. In any case, you can always just create separate tar files by saving each image manually.

I think the docker container driver isn't really an issue because you can download the artifact, load the tar files, and add the image to the build context. I need to double check that  multiplatform doesn't cause issues.

The standard pattern it linked is sort of relevant, but as pointed out, requires a persistent registry. The GitHub actions cache is per-job.

I'll do some more research on proper caching, but to be honest I'm not sure it's worth any amount of complexity in the workflow to save a single build of the smartmontools container, especially when the two builds would happen in parallel.

If the build context works for multiplatform, the easiest thing would be to add an extra build step to each jon in every action..

slydetector:
Taking a step back here, I'm of the opinion having three separate images (omnibus, collector, web) introduces more complexity and friction than they're worth. It would be nice to see a consolidation here but just wanted to throw that out there for future consideration :smile:

kaysond:
> Taking a step back here, I'm of the opinion having three separate images (omnibus, collector, web) introduces more complexity and friction than they're worth. It would be nice to see a consolidation here but just wanted to throw that out there for future consideration 😄

I don't disagree. The omnibus image is, in a way, antithetical to the docker "one thing, one container" philosophy. It still makes sense IMO to separate the collector and web backend, though, as neither depends on the other. Instead of the omnibus image, you'd just run two containers. 


It occurred to me that we're jumping through a lot of hoops to build smartmontools from its own Dockerfile, once, and share it. The collector and omnibus Dockerfiles already duplicate a lot of code. Let's just inline the smartmontools building into both Dockerfiles and then we don't have to touch the workflows at all. 

Longer term we could talk about consolidation of the images or Dockerfiles especially in view of #232

kaysond:
@slydetector - thanks for all your work on this!

slydetector:
> @slydetector - thanks for all your work on this!

Thanks for taking it to the finish line! :smile:

FurkanVG:
Sorry for typing here, do i get it right that the Seagate Drive Command Timeout issue should be fixed? Which :tag would it be? If not, i am sorry. @slydetector @kaysond

kaysond:
> Sorry for typing here, do i get it right that the Seagate Drive Command Timeout issue should be fixed? Which :tag would it be? If not, i am sorry. @slydetector @kaysond

Depends on if your drive is in drivedb, but yes it should. It's in nightly-web and nightly-omnibus. I'm hoping to get one more bug fix in before releasing 0.9.0

FurkanVG:
> > Sorry for typing here, do i get it right that the Seagate Drive Command Timeout issue should be fixed? Which :tag would it be? If not, i am sorry. @slydetector @kaysond
> 
> Depends on if your drive is in drivedb, but yes it should. It's in nightly-web and nightly-omnibus. I'm hoping to get one more bug fix in before releasing 0.9.0

ty, it worked now with master-omnibus, just needed to delete the devices so it got rid of the old error!

kaysond:
> > > Sorry for typing here, do i get it right that the Seagate Drive Command Timeout issue should be fixed? Which :tag would it be? If not, i am sorry. @slydetector @kaysond
> > 
> > 
> > Depends on if your drive is in drivedb, but yes it should. It's in nightly-web and nightly-omnibus. I'm hoping to get one more bug fix in before releasing 0.9.0
> 
> ty, it worked now with master-omnibus, just needed to delete the devices so it got rid of the old error!

FYI master-* is deprecated. New one is nightly-*

SimJoSt:
Check the commit it seems to be released since v0.9.0
However, I am still having the same issue in v0.9.2-omnibus on TrueNAS.
I don't know how to check for drivedb inclusion of these drives: `ST12000VN0007-2GS116`

kaysond:
> Check the commit it seems to be released since v0.9.0 However, I am still having the same issue in v0.9.2-omnibus on TrueNAS. I don't know how to check for drivedb inclusion of these drives: `ST12000VN0007-2GS116`

https://github.com/smartmontools/smartmontools/blob/main/drivedb/drivedb.h

PR #927, opened 2026-02-18
add documentation for rootless systemd service and podman quadlets

defron:
add documentation for rootless systemd service and podman quadlets as discussed in AnalogJ/scrutiny/issues/718

### DIFF
--- a/docs/INSTALL_MANUAL.md
+++ b/docs/INSTALL_MANUAL.md
@@ -168,6 +168,16 @@ Now that we have downloaded the required files, let's prepare the filesystem.
 chmod +x /opt/scrutiny/bin/scrutiny-collector-metrics
 ```
 
+if you are using SELinux, you may need to also do the following:
+
+```sh
+# tell SELinux to allow these binaries
+sudo semanage fcontext -a -t bin_t "/opt/scrutiny/bin(/.*)?"
+# update labels
+sudo restorecon -Rv /opt/scrutiny/bin
+```
+
+
 ### Start Scrutiny Collector, Populate Webapp
 
 Next, we will manually trigger the collector, to populate the Scrutiny dashboard:
--- a/docs/INSTALL_MANUAL.md
+++ b/docs/INSTALL_MANUAL.md
@@ -122,6 +122,11 @@ So you'll need to install the v7+ version using one of the following commands:
     - `dnf install smartmontools`
 - **FreeBSD:** `pkg install smartmontools`
 
+The following additional dependencies are needed if you want to run the collector as an unprivileged user:
+
+- systemd version > 235
+- a restricted user account
+
 ### Directory Structure
 
 Now let's create a directory structure to contain the Scrutiny collector binary.
@@ -133,19 +138,34 @@ mkdir -p /opt/scrutiny/bin
 
 ### Download Files
 
-Next, we'll download the Scrutiny collector binary from the [latest Github release](https://github.com/analogj/scrutiny/releases).
-The file you need to download is named:
+Next, we'll download the Scrutiny collector binary from the [latest Github release](https://github.com/analogj/scrutiny/releases). You are looking for the one titled **scrutiny-collector-metrics-linux-amd64** unless you know you are on arm.
 
-- **scrutiny-collector-metrics-linux-amd64** - save this file to `/opt/scrutiny/bin`
+```sh
+wget -O /tmp/scrutiny-collector-metrics https://github.com/AnalogJ/scrutiny/releases/latest/download/scrutiny-collector-metrics-linux-amd64
+```
+
+Optional, but recommended: Before continuing it's recommended you compare the sha from the release page with the downloaded file to ensure it's the same file and not corrupted/tampered with. The command to do this is:
+
+`echo "SHA_GOES_HERE /tmp/scrutiny-collector-metrics" | sha256sum -c`
+
+example for the v0.8.6 release:
+
+`echo "4c163645ce24e5487f4684a25ec73485d77a82a57f084808ff5aad0c11499ad2 /tmp/scrutiny-collector-metrics" | sha256sum -c`
+
+followed by:
+
+`sudo mv /tmp/scrutiny-collector-metrics /opt/scrutiny/bin/`
+
+to move the binary to its final resting place
 
 
 ### Prepare Scrutiny
 
 Now that we have downloaded the required files, let's prepare the filesystem.
 
-```
+```sh
 # Let's make sure the Scrutiny collector is executable.
-chmod +x /opt/scrutiny/bin/scrutiny-collector-metrics-linux-amd64
+chmod +x /opt/scrutiny/bin/scrutiny-collector-metrics
 ```
 
 ### Start Scrutiny Collector, Populate Webapp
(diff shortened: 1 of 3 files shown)


docs/INSTALL_MANUAL.md
```diff
@@ -154,19 +174,286 @@ Next, we will manually trigger the collector, to populate the Scrutiny dashboard
...
+
+
+#### Creating a Restricted Service Account
+
+This is the account that will run `scrutiny-collector-metrics`. Note this isn't strictly needed for all setups, but is useful from a logging/auditing perspective.
+
+- Debian-based distros:
+    - `sudo adduser --system scrutiny-svc --group --home /opt/scrutiny-svc`
+- RHEL-based distros:
+    - `sudo useradd --system --home-dir /opt/scrutiny-svc --shell /sbin/nologin scrutiny-svc`
```
defron, line 216: full transparency: I haven't tested this on a RHEL clone yet, but it should work. It just took longer than I care to admit to get this all finalized. I'll try to spin up a KVM instance of almalinux or something in the next couple of days. Creating the PR now though because I wanna know if there's any feedback/changes wanted

docs/INSTALL_MANUAL.md
```diff
@@ -154,19 +174,286 @@ Next, we will manually trigger the collector, to populate the Scrutiny dashboard
...
+IPAddressAllow=localhost
+
+[Install]
+WantedBy=multi-user.target
+
+```
+
+Additionally, for nvme drives you may need to create a udev rule on many systems, as /dev/nvme* is often owned only by root:
+
+##### add udev rule `/etc/udev/rules.d/99-nvme.rules` with contents:
```
defron, line 295: So this is annoying and I do think it's possible to work-around it, though would require code changes

nvme drives are character devices and as such only root has permission to them, with the default permission set being 600 root:root

HOWEVER nvme partitions are block devices with a default permission set of 660 root:disk

I dunno if there's appetite in the code to fall back to the nvme partition if the drive isn't accessible. IIRC there have also been some issues with smartctl when done against a namespace, like this one: https://www.smartmontools.org/ticket/1741

defron:
doc typo, amended commit

docs/INSTALL_MANUAL.md
```diff
@@ -154,19 +174,286 @@ Next, we will manually trigger the collector, to populate the Scrutiny dashboard
...
+KeyringMode=private
+LockPersonality=yes
+MemoryDenyWriteExecute=yes
+ProtectSystem=strict
+ProtectHome=yes
+PrivateDevices=no
+## alternatively you can restrict devices using:
+#DevicePolicy=closed
+#DeviceAllow=/dev/sda rw
+#DeviceAllow=/dev/nvme0 rw
```
kaysond: does smartctl really need write access to the raw drives?

docs/INSTALL_MANUAL.md
kaysond: Looks like that fix was merged in 2024, before 7.5 release (which is what we're going to move to shortly). https://github.com/smartmontools/smartmontools/issues/217

An alternative is to add a note for people to specify a namespace (e.g.`nvme0n1`) in the config file... But I do still like the udev rule since it lets people avoid a config file altogether. We should share a command to trigger the rule though:

```sh
sudo udevadm control --reload rules
sudo udevadm trigger --subsystem=block --action=add
```
(from memory; not tested)

docs/INSTALL_MANUAL.md
```diff
@@ -154,19 +174,286 @@ Next, we will manually trigger the collector, to populate the Scrutiny dashboard
...
+After=network.target
+
+[Service]
+Type=oneshot
+User=scrutiny-svc
+Environment="PATH=/opt/smartctl-shim/bin:/usr/bin:/bin"
+ExecStart=/opt/scrutiny/bin/scrutiny-collector-metrics run --api-endpoint "http://localhost:8080"
+
+# --- PRIVILEGE LOCKDOWN ---
+## we use sudo in elevate privileges for smartctl only, so no Ambient Capabilities are needed
```
kaysond: nit:

```suggestion
## we use sudo to elevate privileges for smartctl only, so no Ambient Capabilities are needed
```

docs/INSTALL_ROOTLESS_PODMAN.md
```diff
@@ -0,0 +1,170 @@
...
+
+# create a directory
+sudo mkdir -p /etc/containers/systemd/users/$(id -u scrutiny-svc)
+
+## go into the directory you just created for the rest of the guide
+cd /etc/containers/systemd/users/$(id -u scrutiny-svc)
+```
+
+
+### Qaudlet Files
```
kaysond:
```suggestion
### Quadlet Files
```

kaysond (changes requested):
Thanks for the contribution! Looking good.

kaysond:
I did have one other thought - what about using the setuid bit on smartctl to allow scrutiny to run it elevated without needing a shim?

You'd need to give it permissions like `rwsr-x---` and a custom group for it, say `smartctl`. Then you could do

```
DynamicUser=yes
Group=smartctl
```

which would let smartctl run elevated but wouldnt give scrutiny access beyond that...

I'm also open to changes to scrutiny to help make this more secure...

docs/INSTALL_MANUAL.md
defron: Yup, for the same reason the udev rule is needed. It's a hard requirement for smartctl. Read is obviously needed for reading the smart data from the device. May be able to get away without write but will need to test. Normally write is needed for updating smart data, but also for many other smartctl features related to firmware and raw data modifications

Since scrutiny doesn't use those features it might be possible to get away without write at least, in which case I can also change the udev rule too. 

Will test tonight, as that'd be a nice improvement to my setup too

docs/INSTALL_MANUAL.md
defron: Thanks! I thought I got all the typos yesterday but missed this. Will fix

docs/INSTALL_ROOTLESS_PODMAN.md
defron: Note to self: add quadlet to my qmk firmware autocorrect 🤣 thanks for catching! Will fix tonight

defron:
@kaysond yeah you can set the setuid bit, but it still requires NoNewPrivileges = no so I didn't bother documenting it. I can add that as a third option.

I may be honestly too attached to the sudo way which may just be me being too stubborn and old-school 😅

There's honestly not too many changes I can think of for scrutiny to make, the security implications are pretty much all related to requirements from smartctl. While I haven't done a thorough audit of the scrutiny code, I can't think of anything at the top of my head that you could do differently and your collector is honestly playing extremely nice with very restrictive sandboxing from systemd (besides the aforementioned holes needed for smartctl to work). Most of my issues getting this working has been flubbing a permission or config typo

docs/INSTALL_MANUAL.md
defron: my bad! btw because the issue is for nvme and nvme drives aren't considered block drives, that target won't work but it's basically the same idea. Will have this in my next commit as well as the typo fixes and permission changes

defron:
@kaysond revisiting this:

> I did have one other thought - what about using the setuid bit on smartctl to allow scrutiny to run it elevated without needing a shim?

How much do you want to see this? Thinking over it more I'm hesitant to put it in the docs because while it will work, any update to the smartctl binary for the system will blow it out, which could lead to an increase in the support tickets for it. The advantage for ambientcapabilities + udev rule and sudo + shim is they are one-time efforts that won't get affected by system updates.

If you're really in favor of adding it though I will, though I'll probably instead use `setcap` instead so it doesn't need full root

docs/INSTALL_MANUAL.md
defron: Confirmed for scrutiny it works with just read since it's not initiating a test and updated removing the write permission from this as well as the udev rule

docs/INSTALL_MANUAL.md
defron: commands added in latest commit

docs/INSTALL_MANUAL.md
defron: Tested, worked, though for the collector I needed to add a SELinux label, so included instructions for that

defron: Note: the one thing I wasn't able to test on AlmaLinux was the nvme stuff, but that *should* be the same of just needing the udev rule. I just don't have a spare nvme to pass thru and test it with (I did pass thru a sata drive though and that worked)

kaysond (approved):

kaysond:
The issue with smartctl perms getting blown away on package update did occur to me. I'm happy to merge this as is.

I did come up with another idea, but couldn't quite flesh it out in a way that would be cleaner/simpler than your sudo implementation. You could set up smartctl to run as a socket-activated systemd service where the stdout is piped by systemd to the socket. You could then set that up with whatever permissions are necessary for smartctl only.

Similarly, you would have scrutiny run on a timer and call smartctl via a shim that opens a connection on the socket then rewrites the socket output to its own stdout which scrutiny reads. Scrutiny can then run as a service with no capabilities whatsoever.

The thing I couldn't figure out is how to cleanly pass the smartctl arguments. You could write them over the socket or dump them to a file, but then you'd need a shim on the smartctl service too to read them and actually call smarctl.

defron:
I'm more than happy to have this merged in as-is then but continue discussing with you additional options.

I think if we wanted to explore something further, the way I would suggest doing it would be to have the collector take a smartctl-formatted JSON payload as stdin at which point you could have a separate systemd process whose only responsibility is calling smartctl and then piping it to a socket for a systemd-managed service running the collector using socket activation

With that approach the collector can basically be completely unprivileged and people would be free to pipe data to it however they like (cron or however), and it'd be trivial to write a systemd unit file and timer to regularly pipe data over to it that does the smartctl part.

Systemd really gets you 99% of the way there

I also understand though that we are basically catering to a small minority that want to harden things to the extreme out of paranoia of a supply chain attack XD. I think stdin support on the collector would be the smallest amount of work that would help the few that care enough to set it up to have everything rootless

defron:
gah, right while falling asleep I realized I copied this from my own quadlet and I always do exact-version matches, let me update that to latest tag instead

kaysond:
Thanks for the contribution!

kaysond:
> I think if we wanted to explore something further, the way I would suggest doing it would be to have the collector take a smartctl-formatted JSON payload as stdin at which point you could have a separate systemd process whose only responsibility is calling smartctl and then piping it to a socket for a systemd-managed service running the collector using socket activation

Unfortunately that doesn't work great with the collector architecture as is. The collector runs `smartctl --scan`, then `smartctl --info` for every drive it finds, uploads that to the backend (to check if its archived, among other things), then runs `smartctl --xall` on all the non-archived drives and uploads that separately. It would mean, at minimum, having a separate systemd service for smartctl of each drive.

Another option would be to set up 3 systemd services: smartctl --scan, smartctl --info, and smartctl --xall. The latter two could be templates where the drive name is the template, then you could have one socket per drive name. It means you'd have 3 service files and 2x sockets * # drives, and scan becomes essentially pointless.

I'm personally leaning towards two socket shims where the first writes the arguments and reads the response, and the second reads the arguments and writes the response. I think they'd be trivial to write.

defron:
Do you want to continue in a discussion thread?

I'm actually thinking it should be possible with two systemd services and a single shim and a single socket. Maybe we are saying the same thing just differently? maybe the script for the second systemd service for smartctl is what you mean by the second socket shim?

I'm thinking:

1. systemd unit running the collector
2. simple unix socket for IPC (protected by a user/group)
3. smartctl shim script using socat to talk to the socket (something like `printf "--info\n--json \n" | socat - UNIX-CONNECT:/var/scrutiny/smartctl.sock`)
4. systemd unit using socket activation that kicks off a shell script reading in what type of smartctl command to kick off followed by the paramters

3 and 4 would be simple shell scripts that would be very easy to audit (or for someone to recreate) and would allow the binary for the collector to be fully untrusted

PR #947, opened 2026-03-01
update rootless collector instructions replacing sudo-shim approach with IPC via unix socket

defron:
update rootless collector instructions replacing sudo-shim approach with IPC via unix socket

Also move user creation instructions to podman install as they are no longer needed by the rootless collector

Also fix podman section heading (was nested under Docker, should be at the same level as Manual Installation (without-Docker))

Thoroughly tested on both a Debian-compatible and RHEL-compatible distro

### DIFF
--- a/docs/INSTALL_MANUAL.md
+++ b/docs/INSTALL_MANUAL.md
@@ -315,7 +315,6 @@ NOTE: These cons basically only apply if a major supply-chain attack happens aga
 
 - CAP_SYS_RAWIO allows for data exfiltration/modification from SATA drives (ssh keys, /etc/shadow, etc)
 - CAP_SYS_ADMIN would theoretically allow for significant system compromise
-- nvme drives requires a udev rule for reliable access
 
 
 If you are happy with that, you can jump to [Create a Systemd Timer to run scrutiny-collector.service](#create-a-systemd-timer-to-run-scrutiny-collectorservice)
--- a/docs/INSTALL_MANUAL.md
+++ b/docs/INSTALL_MANUAL.md
@@ -347,7 +347,7 @@ smartctl-shim 0.3
 Args: "${ARGS}"
 Message: ${MSG}
 Please check journald for proxy errors
-and review the docs: 
+and review the docs: https://github.com/AnalogJ/scrutiny/blob/master/docs/INSTALL_MANUAL.md
 EOF
     exit $EXIT_CODE
 }
--- a/docs/INSTALL_MANUAL.md
+++ b/docs/INSTALL_MANUAL.md
@@ -205,20 +205,31 @@ sudo crontab -e
 
 Alternatively you can run `scrutiny-collector-metrics` as non-root so long as the relevant capabilities and permissions are granted.
 
+You have two options for running the collector rootless. An easy setup as well as a more advanced one that thoroughly locks everything down. In both approaches you will need to create a group and udev rule:
 
-#### Creating a Restricted Service Account
 
-This is the account that will run `scrutiny-collector-metrics`. Note this isn't strictly needed for all setups, but is useful from a logging/auditing perspective.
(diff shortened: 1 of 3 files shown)


docs/INSTALL_MANUAL.md
```diff
@@ -325,99 +321,397 @@ NOTE: These cons basically only apply if a major supply-chain attack happens aga
...
+
+```bash
+#!/usr/bin/env bash
+
+# Shim for accounts to use smartctl without being root via unix socket
+
+# don't send json, so it won't be processed by scrutiny
+# but make the error informative for logs
+send_smartctl_shim_error() {
+    local MSG="$1" ARGS="$2" EXIT_CODE="${3:-3}"
```
defron, line 341: I found 3 to be the best exit code in my testing for proxy/socket-related errors.

While yes, smartctl exit codes are a bit-mask. exit code 1 and exit code 2 should be mutually exclusive: 1 is for a command-line parsing error and 2 is for a device access error. If 1 happens, then smartctl never gets to actually check device access, and if 2 happens, then the command-line parsed successfully.

docs/INSTALL_MANUAL.md
```diff
@@ -325,99 +321,397 @@ NOTE: These cons basically only apply if a major supply-chain attack happens aga
...
+fi
+
+# if no exit_status, assume bad.
+# If you run into issues with grepping, switch to jq
+#EXIT_STATUS=$(echo "$RESPONSE" | jq ".smartctl.exit_status // 2" 2>/dev/null)
+EXIT_STATUS=$(echo "$RESPONSE" | grep -m 1 -oP $'[\'"]exit_status[\'"]:\s*\K\d+')
+PARSE_STATUS=$?
+
+# Invalid JSON (smartctl-proxy error) or smartctl error that shouldn't be reported up as json
+if [[ ($PARSE_STATUS -gt 0) || ($EXIT_STATUS -gt 0 && $EXIT_STATUS -lt 4) ]]; then
```
defron, line 369: also in my testing, I had a case where --xall would fail with status_code=4 due to:

"error_recovery_control_supported": false

when using an external drive passed through to a VM (I used PCI-E passthru) due to the USB SATA adapter not supporting it.

The rest of the drive data was valid so 4+ needs to be processed by scrutiny for useful information. I don't feel there's a need to do bitmasking given that, though I am open to discussing if you feel there's an advantage to that

docs/INSTALL_MANUAL.md
```diff
@@ -325,99 +321,397 @@ NOTE: These cons basically only apply if a major supply-chain attack happens aga
...
 ```
 
-3. create a new `scrutiny-collector` file in `/etc/sudoers.d/`
-4. inside `/etc/sudoers.d/scrutiny-collector` add the following:
+4. make the shim executable `sudo chmod +x /opt/smartctl-shim/bin/smartctl`
+5. create a new directory `sudo mkdir -p /opt/smartctl-proxy/bin`
+6. create inside there a proxy script. See below examples using a bash script and python script
+
+
+<details><summary>Click to expand: smartctl-proxy.sh</summary>
```
defron, line 383: both of the sections for the proxy script examples collapse, didn't know github supported this until I was looking for a good way to deal with the example proxy code embeddings

You can see it in action here: https://github.com/defron/scrutiny/blob/chore/rootless-unix-socket-proxy-docs/docs/INSTALL_MANUAL.md#creating-a-restricted-systemd-service-using-smartctl-proxy-via-unix-socket-advanced

kaysond:
@defron - thanks again for this. I'm way behind on taking a look; trying to get some long-overdue bug fixes in before moving on to enhancements

defron:
@kaysond  no worries! Life happens and this is just docs for the most part. On the plus side, I've been running this setup on two servers since creating this PR with no issues so far

PR #973, opened 2026-04-02
Update image tags in example docker compose files

### LINKED ISSUE #967: Docker compose examples have outdated image tags

kaysond:
Fixes #967

### DIFF
--- a/docker/example.hubspoke.docker-compose.yml
+++ b/docker/example.hubspoke.docker-compose.yml
@@ -17,7 +17,7 @@ services:
 
   web:
     restart: unless-stopped
-    image: 'ghcr.io/analogj/scrutiny:master-web'
+    image: 'ghcr.io/analogj/scrutiny:nightly-web'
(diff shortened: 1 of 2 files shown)


hashimaziz1:
Hi, thanks for the work on this package. Just wondering, is there any particular reason for this change, since as far as I understand nightly releases are generally considered less stable than latest.

kaysond:
> Hi, thanks for the work on this package. Just wondering, is there any particular reason for this change, since as far as I understand nightly releases are generally considered less stable than latest.

The example files previously had `master-*` image tags which were renamed to `nightly-*` in the workflow file and everywhere else, but I missed these files. 

You raise a fair point, though our nightlies are pretty stable, and I dislike recommending `latest-*` because people should be pinning to a specific version anyways. Might see if the release workflow can easily be updated to bump versions in the example files. Feel free to open an issue and/or PR. Help is always appreciated.

PR #975, opened 2026-04-04
Friendly Name

newhinton:
Closes #926

First working draft on friendly names.

The names are provided via ENV:
`export FRIENDLY_NAME_123_456_abc_def_12ab=Rivendell`
this corresponds to deviceUUID, except all dashes are now underscores to make it work with the env.

The ui shows either the friendly name, or the device name as a fallback.

### DIFF
--- a/webapp/backend/pkg/notify/notify.go
+++ b/webapp/backend/pkg/notify/notify.go
@@ -20,9 +20,9 @@ import (
 	"github.com/analogj/scrutiny/webapp/backend/pkg/models/measurements"
 	"github.com/analogj/scrutiny/webapp/backend/pkg/thresholds"
 	"github.com/gin-gonic/gin"
+	"github.com/gofrs/uuid/v5"
 	"github.com/nicholas-fedor/shoutrrr"
 	shoutrrrTypes "github.com/nicholas-fedor/shoutrrr/pkg/types"
-	"github.com/gofrs/uuid/v5"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/sync/errgroup"
 )
@@ -127,6 +127,7 @@ type Payload struct {
 	HostId       string `json:"host_id,omitempty"` //host id (optional)
 	DeviceType   string `json:"device_type"`       //ATA/SCSI/NVMe
 	DeviceName   string `json:"device_name"`       //dev/sda
+	FriendlyName string `json:"friendly_name"`     //dev/sda
 	DeviceSerial string `json:"device_serial"`     //WDDJ324KSO
 	Test         bool   `json:"test"`              // false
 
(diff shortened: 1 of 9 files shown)


webapp/backend/pkg/notify/notify.go
```diff
@@ -127,6 +127,7 @@ type Payload struct {
 	HostId       string `json:"host_id,omitempty"` //host id (optional)
 	DeviceType   string `json:"device_type"`       //ATA/SCSI/NVMe
 	DeviceName   string `json:"device_name"`       //dev/sda
+	FriendlyName string `json:"friendly_name"`     //dev/sda
```
kaysond, line 130: nit: comment should be changed

webapp/backend/pkg/notify/notify.go
```diff
@@ -20,9 +20,9 @@ import (
 	"github.com/analogj/scrutiny/webapp/backend/pkg/models/measurements"
 	"github.com/analogj/scrutiny/webapp/backend/pkg/thresholds"
 	"github.com/gin-gonic/gin"
+	"github.com/gofrs/uuid/v5"
```
kaysond, line 23: doesn't look like this is used in the file?

kaysond (changes requested):
@newhinton - thanks for taking this on!

I think it would be better to store the friendly name in the db rather than depend on env vars. The env var is kinda messy, and if that's the method of setting it, we should also add support to it for in the config file, which would make things more complicated. Organizationally, adding it to the db is the more logical place, and it's easier for the user IMO.

This would entail:
* add a migration to webapp/backend/pkg/database/scrutiny_repository_migrations.go
  * add the device model to a dir in webapp/backend/pkg/database/migrations
  * use automigrate like the m20250221084400 migration
* add a POST route for something like /device/:scrutiny_uuid/friendly_name in webapp/backend/pkg/web/server.go
* add a handler in webapp/backend/pkg/web/handler to set the name in the db
* maybe update mock data - `mockgen -source=webapp/backend/pkg/database/interface.go -destination=webapp/backend/pkg/database/mock/mock_database.go`
* add something in the UI to set it. easiest might be adding a menu item in webapp/frontend/src/app/layout/common/dashboard-device/dashboard-device.component.html

Also, regardless of the backend setup, mock data in the frontend needs to be updated to add a friendly name.

I've been meaning to remove the host id from webapp/frontend/src/app/shared/device-title.pipe.ts (in deviceTitleWithFallback) because it's shown above too. Maybe we could throw that in here.

Please let me know if you have any questions or if there's anything I can do to help.

newhinton:
I will try my hand at your suggested implementation!

Though it will take a while, i am not familiar with go. If i get stuck, ill ask for help. Though your action-plan looks like a thorough guide!

PR #1010, opened 2026-05-20
Fix smart support detection

### LINKED ISSUE #1004: smart_support is no longer a boolean
See https://github.com/search?q=repo%3AAnalogJ%2Fscrutiny%20smart_support&type=code
and https://github.com/AnalogJ/scrutiny/discussions/1001#discussioncomment-16721243

Low prio as it isnt actually used anywhere

puneetdixit200:
## Summary

- Parse `smart_support` from smartctl metadata when it is reported as either a boolean or the newer `{ "available": ..., "enabled": ... }` object.
- Copy the parsed support value into the collector device registration payload.
- Add a regression test using the smartctl object shape from the linked discussion.

Fixes #1004

## Validation

Passed:

- `go test ./collector/pkg/detect -run 'TestDetect_SmartCtlInfo/should_report_smart_support_from_object' -count=1`
- `go test ./collector/pkg/detect ./webapp/backend/pkg/models/collector -count=1`
- `go test ./collector/... ./webapp/backend/pkg/models/... -count=1`
- `git diff --check`

Also ran `go test ./...`; all collector and backend model packages passed, but the full run failed in `webapp/backend/pkg/web` because local InfluxDB is not running on `localhost:8086`.

I also ran the repo lint command. The Makefile target first failed because `~/go/bin` was not on `PATH`; after running the installed linter directly, it reported existing repository-wide findings outside this change, including errcheck/staticcheck/gofmt issues in unrelated files. No lint findings were reported for the files touched here.

## AI usage

Tool used: OpenAI GPT-5.

AI assistance was used to trace the collector parsing path, draft the small code/test change, and prepare this PR text. The final diff was reviewed and validated with the commands above. The behavioral verification is at the smartctl JSON parsing level using the `smart_support` object shape from the accepted issue/discussion; no Synology-specific hardware behavior is changed here.

## Review follow-up validation

2026-05-21:

- `git diff --check`
- `go test ./collector/pkg/detect -run 'TestDetect_SmartCtlInfo' -count=1`
- `go test ./webapp/backend/pkg/models/collector -count=1`

A broader local Windows run of `go test ./collector/pkg/detect ./webapp/backend/pkg/models/collector -count=1` still fails in existing scan tests that expect names like `sda` while the current fixtures produce `/dev/sda`. The touched `SmartCtlInfo` path and backend collector model package passed.

### DIFF
--- a/collector/pkg/detect/detect_test.go
+++ b/collector/pkg/detect/detect_test.go
@@ -357,9 +357,11 @@ func TestDetect_SmartCtlInfo(t *testing.T) {
 			someCapacity       int64 = 3840755982336
 		)
 
+		fullDeviceName := detect.DevicePrefix() + someDeviceName
+
 		fakeConfig := mock_config.NewMockInterface(ctrl)
 		fakeConfig.EXPECT().
-			GetCommandMetricsInfoArgs("/dev/" + someDeviceName).
+			GetCommandMetricsInfoArgs(fullDeviceName).
 			Return(someArgs)
 		fakeConfig.EXPECT().
 			GetString("commands.metrics_smartctl_bin").
@@ -372,7 +374,7 @@ func TestDetect_SmartCtlInfo(t *testing.T) {
 
 		fakeShell := mock_shell.NewMockInterface(ctrl)
 		fakeShell.EXPECT().
-			Command(someLogger, "smartctl", append(strings.Split(someArgs, " "), "/dev/"+someDeviceName), "", gomock.Any()).
+			Command(someLogger, "smartctl", append(strings.Split(someArgs, " "), fullDeviceName), "", gomock.Any()).
 			Return(string(smartctlInfoResults), err)
 
 		d := detect.Detect{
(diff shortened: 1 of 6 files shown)


collector/pkg/detect/detect_test.go
```diff
@@ -396,4 +396,72 @@ func TestDetect_SmartCtlInfo(t *testing.T) {
...
+		fakeConfig.EXPECT().
+			GetCommandMetricsInfoArgs("/dev/" + someDeviceName).
+			Return(someArgs)
+		fakeConfig.EXPECT().
+			GetString("commands.metrics_smartctl_bin").
+			Return("smartctl")
+
+		someLogger := logrus.WithFields(logrus.Fields{})
+
+		smartctlInfoResults := `{
```
kaysond: this should be in a testdata file. see the other tests in this file.

we should have a test for both the old boolean and new object

webapp/backend/pkg/models/collector/smart.go
```diff
@@ -241,6 +244,35 @@ type SmartInfo struct {
...
+	Enabled   bool `json:"enabled"`
+}
+
+func (s *SmartSupport) UnmarshalJSON(data []byte) error {
+	var supported bool
+	if err := json.Unmarshal(data, &supported); err == nil {
+		s.Available = supported
+		s.Enabled = supported
+		return nil
+	}
```
kaysond, line 261: please add a comment explaining why this is needed

kaysond (changes requested):

collector/pkg/detect/detect_test.go
puneetdixit200: Moved the smartctl JSON into `collector/pkg/detect/testdata` and added cases for both the legacy boolean and object forms in `e19f722`.

webapp/backend/pkg/models/collector/smart.go
puneetdixit200: Added a compatibility comment above `SmartSupport.UnmarshalJSON` in `e19f722` explaining the old boolean form and newer available/enabled object form.

kaysond (approved):
Thanks for the contribution!

PR #1024, opened 2026-06-13
Fix device failed status not getting cleared when SMART data becomes clean

kaysond:
Also remove a redundant setting of the status bit

### DIFF
--- a/webapp/backend/pkg/database/scrutiny_repository_device.go
+++ b/webapp/backend/pkg/database/scrutiny_repository_device.go
@@ -54,14 +54,18 @@ func (sr *scrutinyRepository) UpdateDevice(ctx context.Context, scrutiny_uuid uu
 }
 
 // Update Device Status
+// The provided status reflects the latest full SMART analysis and replaces the previous value
+// outright, so a device that recovers (clean SMART data) is reset from failed back to passing.
 func (sr *scrutinyRepository) UpdateDeviceStatus(ctx context.Context, scrutiny_uuid uuid.UUID, status pkg.DeviceStatus) (models.Device, error) {
 	var device models.Device
 	if err := sr.gormClient.WithContext(ctx).Where("scrutiny_uuid = ?", scrutiny_uuid.String()).First(&device).Error; err != nil {
 		return device, fmt.Errorf("could not get device from DB: %v", err)
 	}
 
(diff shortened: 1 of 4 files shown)


kaysond:
@rbock44 - do you still have your old sqlite db backed up? Could you test out this fix?

rbock44:
Unfortunately not. But I can try to reproduce it as I can remove the wrapper script around smartctl which manipulates the Command_Timout values. So the failed status should come again.

By accident I read a summary of an AI about the failed status and it said that the policy is once failed status is not reset. Which opens the question is it really a bug or a feature :-) Lets say some policies change and a drive is now green which was red before I would want a feature to reset the status in the UI or via API but not to automatically turn green. Well that is my flavour.

Not sure what the fix is your are done, but I suspect you automatically reset the status to green ...

Just let me know if I should reproduce the problem or if I should create a feature request instead :-)

Thanks for your god work any way!

kaysond:
> Unfortunately not. But I can try to reproduce it as I can remove the wrapper script around smartctl which manipulates the Command_Timout values. So the failed status should come again.
> 
> By accident I read a summary of an AI about the failed status and it said that the policy is once failed status is not reset. Which opens the question is it really a bug or a feature :-) Lets say some policies change and a drive is now green which was red before I would want a feature to reset the status in the UI or via API but not to automatically turn green. Well that is my flavour.
> 
> Not sure what the fix is your are done, but I suspect you automatically reset the status to green ...
> 
> Just let me know if I should reproduce the problem or if I should create a feature request instead :-)
> 
> Thanks for your god work any way!

It's probably talking about this: https://github.com/AnalogJ/scrutiny/issues/374#issuecomment-1257031291

@AnalogJ - curious what your rationale was behind this behavior. I can vaguely understand why you'd want a "sticky" failure, but I think it causes problems with transient behavior (e.g. temporary failures with bad cables, or the transient ECC in #374) and smartmontools issues like the seagate timeout, where the data starts out bad because the arguments to correctly parse the timeout int aren't there, then the user fixes them but the drive stays failed.

An alternative is to have a reset option in the GUI, but IMO it's fine to have the drive state in scrutiny match whatever is currently reported by smartmontools, even if there was a failure in the past. If an attribute changes again, scrutiny will still then mark the failure and notify the user...

Issue #945, opened 2026-02-27
Add notifications for collector failures


kaysond:
We should add an API endpoint so the collector can report `smartctl --scan` and `smartctl --info` errors. The endpoint can trigger an opt-out notification. If the backend receives a `smartctl --xall` json object indicating an error (see #944), a notification should also be sent.

This needs a new [handler](https://github.com/AnalogJ/scrutiny/tree/master/webapp/backend/pkg/web/handler), a new route:

https://github.com/AnalogJ/scrutiny/blob/9d1ce790d0cad95820f8148af272114f7f5c22a9/webapp/backend/pkg/web/server.go#L22-L53

and a new setting, including a UI element and database migration. See https://github.com/AnalogJ/scrutiny/commit/bdbe13e320673d017c9c5cf3aea23436c79642c4 for an example.

kaysond:
This probably needs a special case for #942

PR #1029, opened 2026-07-11
fix(web): skip persisting SMART metrics when smartctl could not read the device

### LINKED ISSUE #944: Backend will add invalid data to the database if smartctl returns a json-formatted error
smartctl will return a non-zero exit code under a number of conditions, defined in their source [here](https://github.com/smartmontools/smartmontools/blob/b1bb7d73c37c16ddddf73906ac9e9e9ad673d481/src/smartctl.h#L16-L47).

If the collector gets a non-zero exit code for a `smartctl --xall`, it will still send the data to the backend:

https://github.com/AnalogJ/scrutiny/blob/9d1ce790d0cad95820f8148af272114f7f5c22a9/collector/pkg/collector/metrics.go#L136-L143

which will happily unmarshall it into our model and push it (with a bunch of uninitialized struct fields) into the database:

https://github.com/AnalogJ/scrutiny/blob/9d1ce790d0cad95820f8148af272114f7f5c22a9/webapp/backend/pkg/web/handler/upload_device_metrics.go#L45-L51

We should update the `exit_status` field in our model to match the bitmask from the smartctl source:

https://github.com/AnalogJ/scrutiny/blob/9d1ce790d0cad95820f8148af272114f7f5c22a9/webapp/backend/pkg/models/collector/smart.go#L1-L11

and in the backend handler shown above, skip pushing data if any of bits 2-0 are set.

Note that some error codes are shared! Smartctl exits 2 for a variety of reasons including a drive being in standby.

TowyTowy:
Fixes #944

smartctl reports its result as an exit-status bitmask. When bit 0 (command line did not parse) or bit 1 (device open failed / low-power mode) is set, the JSON payload is largely uninitialized — empty model/serial, no attributes, and a zero-value `smart_status` that the backend reads as a SMART failure. `UploadDeviceMetrics` persisted this unconditionally, so a drive that was merely in standby or couldn't be opened got stored as a bogus *failing* device (and could even trigger failure notifications).

**Change:** add `SmartInfo.HasInvalidData()`, which inspects the smartctl exit-status bitmask, and skip persisting the payload in the upload handler when it's set.

**Note on bit 2:** the issue suggested skipping on bits 0-2, but I scoped this to bits 0-1 and deliberately excluded bit 2 (0x04 — checksum / SMART-command error). RAID/megaraid passthrough sets bit 2 while still returning valid data (`testdata/smart-megaraid0.json`, exit_status 4: full model + 20 attributes + health), so dropping it would regress those drives. Happy to widen the mask if you'd prefer — flagging it for your call.

**Testing:** unit tests over the exit-status bitmask (synthetic values + `smart-ata`/`smart-fail`/`smart-fail2`/`smart-megaraid0` fixtures) and an end-to-end handler test (real InfluxDB) asserting an exit-2 payload yields zero stored `smart_results`. `go test ./webapp/backend/pkg/models/collector/... ./webapp/backend/pkg/web/...` passes; `gofmt`/`go vet` clean. I do not have megaraid/RAID hardware, so that path was preserved rather than modified and not hardware-tested.

**AI disclosure (per `AI_POLICY.md`):** implemented with AI assistance (Claude Code). All code and tests were reviewed and verified by me by running the tests locally against a real InfluxDB instance; the bit-2/megaraid scoping decision was made deliberately based on the committed fixtures.

### DIFF
--- a/webapp/backend/pkg/models/collector/smart.go
+++ b/webapp/backend/pkg/models/collector/smart.go
@@ -287,6 +287,44 @@ func (s *SmartInfo) Capacity() int64 {
 	return 0
 }
 
+// smartctl reports its result as an exit status bitmask (see smartmontools'
+// smartctl.h). The individual bits below indicate why smartctl set a non-zero
+// exit status:
+//
+//	bit 0 (0x01): the command line did not parse
+//	bit 1 (0x02): the device could not be opened, or is in a low-power/standby mode
+//	bit 2 (0x04): a SMART or ATA command to the device failed, or a checksum
+//	              error was found in a SMART data structure
+//
+// The remaining bits (3-7) report genuine health findings (SMART "FAILING"
+// status, pre-fail attributes, logged errors, etc.) whose data we still want
+// to keep, so they are intentionally not part of the mask below.
+const (
+	smartctlExitStatusCommandLineDidNotParse = 1 << 0 // 0x01
+	smartctlExitStatusDeviceOpenFailed       = 1 << 1 // 0x02
+)
+
+// smartctlExitStatusDataUnavailableMask covers the exit status bits that mean
+// smartctl could not actually read usable data off the device. When either is
+// set the JSON payload is largely uninitialized (empty model/serial, no
+// attributes, and a zero-value smart_status that looks like a SMART failure),
+// so it must not be persisted.
+//
+// Note that bit 2 (0x04) is deliberately excluded: RAID/megaraid passthrough
+// frequently sets it (smartctl cannot read some log pages through the
+// controller) while still returning a complete, valid identity + attribute set
+// - see testdata/smart-megaraid0.json (exit_status 4). Dropping that data would
+// regress monitoring for those drives.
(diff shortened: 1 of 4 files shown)


kaysond:
Hi there. Thanks for the contribution. I've actually been working on this myself, though progress has stalled lately. Unfortunately this issue is a bit more complicated than what you've submitted. (e.g. see https://github.com/AnalogJ/scrutiny/discussions/942). My plan was to ship a fix for this and #945 in a single PR, since they're related.

If you'd like to work on this, can you please take a look at https://github.com/AnalogJ/scrutiny/tree/collector_error_handling as a starting point? It's mostly there, but needs implementation of the notifications and obviously more testing.

TowyTowy:
Thanks for the context, and no worries at all — your combined approach on `collector_error_handling` (handling this together with #945, plus notifications) is clearly the more complete fix, so I'll defer to that. I'll close this one so it's not in your way. If a specific piece would be useful for me to pick up on that branch, happy to help — just point me at it.

PR #1038, opened 2026-07-27
Update frontend dependencies

kaysond:
Closes #937

Refreshes the frontend dependency tree so the open dependabot PRs can be closed, and drops the
packages that were pulling in vulnerabilities without being used. No UI changes.

## What changed

**Removed unused packages** (each verified to have no reachable usage in `src/`):

| Package | Why it can go |
| --- | --- |
| `crypto-js`, `@types/crypto-js` | zero references anywhere in `src/` or `e2e/` |
| `rrule` | zero references |
| `ngx-markdown` | only `MarkdownModule.forRoot({})` in `app.module.ts`; no `<markdown>` element, directive, component or service is used anywhere. Pulled in `marked`, `prismjs` and `katex`. |
| `quill` | no JS/TS import at all — only `quill.snow.css` in `vendors.scss` and a `.ql-*` override sheet. Nothing in the app renders `.ql-*` markup. |
| `tslint`, `codelyzer`, `tslint.json`, the `lint` target | `angular.json` pointed at `@angular-devkit/build-angular:tslint`, a builder that no longer exists in the installed `@angular-devkit/build-angular@13.3.11`, so `npm run lint` was already failing. |
| `protractor`, `jasmine-spec-reporter`, `@types/jasminewd2`, `e2e/`, the `e2e` target | Protractor is EOL and nothing runs the e2e target (not in CI, not in the `Makefile`). It was the source of `request`, `form-data@2.3.3`, `webdriver-manager`, `adm-zip`, `xml2js` and `tough-cookie`. |
| `@types/highlight.js` | deprecated stub; highlight.js has shipped its own types since v10 |

**Bumped every remaining direct dependency** to the newest release that still works on this
toolchain, and fixed three ranges that were already unsatisfiable — on `master` a plain
`npm install` (as opposed to `npm ci`) resolves them to versions that cannot build:

| Package | Was | Resolved to | Broke because | Now |
| --- | --- | --- | --- | --- |
| `@types/lodash` | `^4.14.188` | 4.17.x | `infer N extends number` needs TypeScript ≥ 4.7 | `~4.14.202` |
| `@types/node` | `^18.11.9` | 18.19.130 | 18.19.48+ uses `typeof ServerResponse<…>` generics, needs TypeScript ≥ 4.7 | `18.19.47` |
| `ng-apexcharts` | `^1.7.4` | 1.8.0 | ships an Angular 14+ `ɵɵComponentDeclaration` (8 type args; Angular 13 wants 6) | `~1.7.7` |

[...]

PR #1039, opened 2026-07-27
Keep the dashboard temperature tooltip inside the chart

### LINKED ISSUE #950: Styling issue: Temperatures out of graph container bounds
### Discussed in https://github.com/AnalogJ/scrutiny/discussions/948

<sup>Originally posted by **probablyjassin** March  2, 2026</sup>
### Issue Description

Temerature data get's cut off and becomes unreadable for multiple disks
image

### Expected Behavior

Should position itself in a way that stays readable

### Actual Behavior

goes off screen

### Reproduction Steps

Add multiple drives and let them measure drive temps for multiple days

### scrutiny logs

```text
n/a
```

### Scrutiny Version

master-omnibus (master#c95b272)

### Smartmontools Version

docker

### OS Version Information

Ubuntu

### Component

web

### Deployment Method

- [X] docker
- [ ] binaries
- [ ] systemd

### Hard Drive Information

_No response_

### smartctl output

_No response_

### scrutiny.yaml

_No response_

### collector.yaml

_No response_

_No response_

### I acknowledge that:

kaysond:
Closes #950

## What was happening

The dashboard temperature chart runs in `sparkline` mode, which forces `legend.show = false`, so the
only way to tell which line is which drive is the shared tooltip. That tooltip grows one row per
device, and two things then conspire against it:

1. ApexCharts only keeps the *top half* of the tooltip inside the chart. `Position.moveTooltip()`
   clamps with `if (tooltipRect.ttHeight / 2 + y > gridHeight)`, so the bottom of the tooltip can
   sit up to `ttHeight / 2` below the chart.
2. The card wrapping the chart is `rounded overflow-hidden`, so whatever hangs below gets clipped.

Measured on a 1280×900 viewport with 7 drives, hovering the series the way the reporter's screenshot
shows: the tooltip is 266px tall and **119px of it is cut off**.

## The fix

- pin the tooltip with `tooltip.fixed = { enabled: true, position: 'topLeft' }`. That skips
  `moveTooltip()` entirely (`if (!ttCtx.fixedTooltip) this.moveTooltip(...)`) and anchors the tooltip
  to the top-left of the chart, so it is laid out downwards from a known point instead of around the
  cursor.
- drop `overflow-hidden` from the chart card. For a large number of drives the tooltip is simply
  taller than the chart, and clipping it is what made it unreadable. The temperature card is the last
  block on the dashboard, so a tall tooltip just extends into empty page space.

Both are needed: `fixed` alone still clips from 8 drives up, because the tooltip becomes taller than
the 275px chart.

## Measurements

Driven with headless Chromium against a production `ng serve`, hovering the chart and measuring the
tooltip against the clipping ancestor. "clipped" is how many pixels of the tooltip were cut off:

| drives | tooltip height | before | `fixed` only | this PR |
| --- | --- | --- | --- | --- |
| 7 | 266px | 119px clipped | 0 | 0 |
| 8 | 299px | clipped | 12px clipped | 0 |
| 12 | 431px | clipped | 144px clipped | 0 |
| 16 | 563px | clipped | 276px clipped | 0 |

Also checked at a 414×896 viewport (7 drives): tooltip fits, nothing clipped.

## Not fixed here

The palette in `_prepareChartData()` has only 6 colours, and ApexCharts recycles them, so from 7
drives up two lines share a colour. That is a separate (and more debatable) change — the tooltip
markers have the same problem, and fixing it properly probably means generating colours rather than
listing them.

## AI disclosure

Per [AI_POLICY.md](AI_POLICY.md): this change was written by **Claude Code** (Opus 5). It reproduced
the bug, produced the numbers above, and made the change. Everything ran in a `node:24-trixie`
container driving headless Chromium via puppeteer against `ng serve --configuration production`, with
the API responses stubbed at the network layer so the drive count could be varied; nothing was
installed on the host.

Automated verification performed:

- reproduced the clipping at 7 drives and confirmed it is gone after the change, at 7/8/12/16 drives
  and at a phone-sized viewport
- `npm run build:prod` and `npx ng test --watch=false --browsers=ChromeHeadless --code-coverage`
  (87/87)

No unit test was added: the change is chart configuration, and a test asserting the option is set
would only restate the diff. The behaviour that matters is geometric and was checked in a browser.

That is the extent of the verification behind this PR as opened.

### DIFF
--- a/webapp/frontend/src/app/modules/dashboard/dashboard.component.html
+++ b/webapp/frontend/src/app/modules/dashboard/dashboard.component.html
@@ -78,7 +78,8 @@
 
             <!-- Drive Temperatures -->
             <div class="flex flex-auto w-full min-w-80 h-90 p-4">
-                <div class="flex flex-col flex-auto bg-card shadow-md rounded overflow-hidden">
+                <!-- deliberately not overflow-hidden: that clips the tooltip once it outgrows the chart -->
+                <div class="flex flex-col flex-auto bg-card shadow-md rounded">
                     <div class="flex flex-col p-6 pr-4 pb-4">
                         <div class="flex items-center justify-between">
                             <div class="flex flex-col">
--- a/webapp/frontend/src/app/modules/dashboard/dashboard.component.spec.ts
+++ b/webapp/frontend/src/app/modules/dashboard/dashboard.component.spec.ts
@@ -1,0 +1,34 @@
+import {tooltipViewportOffset} from './dashboard.component';
+
+describe('tooltipViewportOffset', () => {
+    const viewport = {width: 1000, height: 800};
+    const box = (left: number, top: number, width: number, height: number) =>
+        ({left, top, right: left + width, bottom: top + height});
+
+    it('leaves a tooltip that already fits where it is', () => {
+        expect(tooltipViewportOffset(box(100, 100, 200, 300), viewport)).toEqual({dx: 0, dy: 0});
+    });
+
+    it('lifts a tooltip that runs off the bottom', () => {
+        expect(tooltipViewportOffset(box(100, 600, 200, 300), viewport)).toEqual({dx: 0, dy: -108});
+    });
+
+    it('pulls a tooltip that runs off the right back to the left', () => {
+        expect(tooltipViewportOffset(box(900, 100, 200, 100), viewport)).toEqual({dx: -108, dy: 0});
+    });
+
+    it('pins a tooltip taller than the viewport to the top rather than lifting it out of view', () => {
+        expect(tooltipViewportOffset(box(100, 200, 200, 900), viewport)).toEqual({dx: 0, dy: -192});
+    });
+
+    it('does not push a tooltip off the left while pulling it in from the right', () => {
+        expect(tooltipViewportOffset(box(-50, 100, 1200, 100), viewport)).toEqual({dx: 58, dy: 0});
+    });
+
+    it('settles: re-applying the offset to the moved tooltip is a no-op', () => {
+        const start = box(900, 600, 200, 300);
+        const {dx, dy} = tooltipViewportOffset(start, viewport);
+        const moved = box(start.left + dx, start.top + dy, 200, 300);
+        expect(tooltipViewportOffset(moved, viewport)).toEqual({dx: 0, dy: 0});
+    });
+});
--- a/webapp/frontend/src/app/modules/dashboard/dashboard.component.ts
+++ b/webapp/frontend/src/app/modules/dashboard/dashboard.component.ts
@@ -218,6 +218,12 @@ export class DashboardComponent implements OnInit, AfterViewInit, OnDestroy
                 theme: 'dark',
                 shared: true,
                 intersect: false,
+                // the shared tooltip grows a row per device, and apexcharts only keeps the top half of
+                // it inside the chart. Pinning it to a corner anchors it instead.
+                fixed: {
+                    enabled: true,
+                    position: 'topLeft'
+                },
                 x    : {
                     format: 'MMM dd, yyyy HH:mm:ss'
                 },
--- a/webapp/frontend/src/app/modules/dashboard/dashboard.component.ts
+++ b/webapp/frontend/src/app/modules/dashboard/dashboard.component.ts
@@ -2,6 +2,8 @@ import {
     AfterViewInit,
     ChangeDetectionStrategy,
     Component,
+    ElementRef,
+    NgZone,
     OnDestroy,
     OnInit,
     ViewChild,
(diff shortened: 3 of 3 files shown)


kaysond:
Reworked — the `tooltip.fixed` corner pin is gone. The tooltip now stays where apexcharts puts it and is only moved when it would otherwise run off screen, at which point it goes up and left from the cursor.

## Why there is no option for this, and why bumping does not help

I went looking for the supported way to do this first. There isn't one: the full tooltip option set in 3.37.3 is `enabled / enabledOnSeries / shared / followCursor / intersect / inverseOrder / custom / fillSeriesColor / theme / cssClass / style / fixed / x / y / z / marker / items / onDatasetHover` — nothing about boundaries, collision or flipping.

And the positioning code is unchanged across every published major:

| version | `Position.moveTooltip` vertical clamp |
| --- | --- |
| 3.37.3 (current) | `if (tooltipRect.ttHeight / 2 + y > gridHeight)` |
| 3.54.1 (what #1038 moves to) | identical |
| 4.7.0 | identical |
| 5.10.3 (latest) | identical, `w.globals` renamed to `w.layout` |

So **a dependency bump does not fix this**, which is worth knowing before #1038 or the Angular work lands.

That `/ 2` is the bug: the clamp only fires once the tooltip's *midpoint* passes the bottom of the plot area, so the bottom half is allowed to hang out. And when the tooltip is taller than the plot area the clamp cannot help at all — `y = gridHeight - ttHeight` is negative, so it pushes the tooltip out of the top instead. Since `sparkline` mode forces `legend.show = false`, the shared tooltip is the only way to tell the lines apart, and it grows a row per drive: 7 drives is 266px against a ~284px plot area, 16 drives is 563px. It is simply bigger than the box apexcharts is trying to fit it into.

`followCursor: true` was the closest built-in candidate — it does flip up and left — but it flips at the *midpoint of the grid* rather than when an overflow would actually occur, and it skips the clamp entirely, so a tall tooltip still leaves the chart. It just moves the problem.

## The change

Two parts:

- the card is no longer `overflow-hidden`, which is what was clipping the part that escaped
- a `MutationObserver` on the tooltip's `style` shifts it back inside the viewport after each move apexcharts makes

```ts
export function tooltipViewportOffset(box: Box, viewport: { width: number; height: number },
                                      margin: number = 8): { dx: number; dy: number } {
    // pull it back from the far edges first, then off the near ones, so a tooltip too big to fit
    // lands against the top left rather than being pushed further off screen
    let dx = Math.min(0, viewport.width - margin - box.right);
    let dy = Math.min(0, viewport.height - margin - box.bottom);
    dx += Math.max(0, margin - (box.left + dx));
    dy += Math.max(0, margin - (box.top + dy));

    return {dx, dy};
}
```

Because the correction is a clamp rather than a jump to a fixed anchor, the tooltip does not move at all until it actually would overflow — below 8 drives on a normal window it never fires.

Two things this cost me, both of which are commented in the source because they are not guessable:

**The tooltip animates.** `.apexcharts-tooltip` carries `transition: .15s ease all`, so reading the position back with `getBoundingClientRect()` returns a point it is still travelling away from. My first attempt did exactly that, never converged, and pinned the renderer at 90% CPU until puppeteer gave up with `Input.dispatchMouseEvent timed out`. It now measures the position apexcharts *asked* for (`style.left`/`style.top` plus `offsetWidth`/`offsetHeight`), which is stable, so re-entering the observer after our own write computes `{dx: 0, dy: 0}` and stops. There is a test for that settling property.

**The observer is created inside `runOutsideAngular`.** zone.js patches `MutationObserver`, so without it every tooltip move would trigger a change detection pass on an `OnPush` dashboard.

## Measurements

Headless Chromium against `ng serve --configuration production`, API stubbed at the network layer so the drive count is controllable. 12 hover positions per run (4 across × 3 down), reporting the worst case of each. "clipped" is against the nearest clipping ancestor, "offscreen" is against the viewport.

Before — the card is the clipper:

| drives | tooltip height | clipped | offscreen |
| --- | --- | --- | --- |
| 7 | 266px | 0px | 0px |
| 12 | 431px | **144px** | 96px |
| 16 | 563px | **276px** | 228px |

After — nothing clips it any more, so the clipper is the viewport:

| drives | tooltip height | clipped | offscreen |
| --- | --- | --- | --- |
| 7 | 266px | 0px | 0px |
| 8 | 299px | 0px | 0px |
| 12 | 431px | 0px | 0px |
| 16 | 563px | 0px | 0px |
| 20 | 695px | 0px | 0px |

Also clean at 414×896 (8 and 16 drives).

The one case that still overflows is when the tooltip is taller than the *window*, not just the card. At 1280×**500**:

| drives | tooltip height | offscreen |
| --- | --- | --- |
| 16 | 563px | 71px |
| 20 | 695px | 203px |

That is exactly `ttHeight - 500 + margin` in both rows — the tooltip is pinned to the top of the window and what hangs below is the part that cannot fit anywhere. Nothing can be done about that short of not drawing a row per drive, so I left it; the important part is it clamps to the top instead of lifting out of view, which the unit test covers.

Frontend suite: **93/93** (87 before, 6 new).

Unit test covers the geometry directly, including the case where the tooltip is taller than the window (it pins to the top instead of lifting out of view) and the settling property that the runaway-loop bug violated.

## Not addressed

The 6-colour palette is still recycled, so 7+ drives share line colours. That is the reason a shared tooltip is load-bearing here at all, but it is a separate change.

PR #1040, opened 2026-07-27
Fix release version reporting in frontend builds, the release workflow, and docker images

### LINKED ISSUE #998: Scrutiny releases do not consistently show the correct version
### Discussed in https://github.com/AnalogJ/scrutiny/discussions/993

<sup>Originally posted by **soloam** April 24, 2026</sup>
### Issue Description

This is a small "erro", but even qhen downloading a released version, the version printed is the master#<<commit>> and not the actual version

### Expected Behavior

If I download a release version I would like to see the version so that I can see if I need to update

### Actual Behavior

If I download a release version I would like to see the version so that I can see if I need to update

### Reproduction Steps

1. Install a released version

### scrutiny debug logs

_No response_

### Scrutiny Version

v0.9.1

### Smartmontools Version

7.2

### OS Version Information

Debian 13

### Component

web

### Deployment Method

- [ ] docker
- [X] binaries
- [ ] systemd

### Hard Drive Information

_No response_

### smartctl output

_No response_

### docker-compose.yml

_No response_

### scrutiny.yaml

_No response_

### collector.yaml

_No response_

_No response_

### I acknowledge that:

kaysond:
Closes #998

## What was happening

`webapp/frontend/git.version.sh` derives the version from `GITHUB_REF_NAME` / `GITHUB_REF_TYPE` when
`CI` is set. Two workflows call it, and they land on opposite sides of that check:

- `docker-build.yaml` runs on a **tag** push, so `GITHUB_REF_TYPE=tag` and the version is right.
- `release.yaml` is a **`workflow_dispatch` on master**, so `GITHUB_REF_TYPE=branch` and
  `build_frontend` writes `master#<sha>` into `versions.ts`.

The web UI shows that string, so every `scrutiny-web-frontend.tar.gz` attached to a release reports
the branch. That matches the report: v0.9.1, binaries.

Confirmed against what is actually published:

```
$ curl -sSL https://github.com/AnalogJ/scrutiny/releases/download/v0.9.1/scrutiny-web-frontend.tar.gz | tar -xz
$ grep -o 'appVersion="[^"]*"' dist/main.bf546340fdb53ee0.js
appVersion="master#d77976c"

$ docker run --rm --entrypoint sh ghcr.io/analogj/scrutiny:v0.9.1-web \
    -c 'grep -oh "appVersion=\"[^\"]*\"" /opt/scrutiny/web/*.js'
appVersion="v0.9.1"
```

So the docker images were already fine and only the binary/manual install path is wrong.

## The fix

`git.version.sh` takes an explicit `SCRUTINY_VERSION` ahead of the existing logic, and `release.yaml`
passes the version that `packagrio/action-bumpr-go` has just written to `version.go` — which is the
authoritative version for that release, and is present in the workspace artifact the frontend job
downloads. The step fails the job rather than silently producing a bad string if it cannot read it.

Reading `version.go` avoids depending on the action's step output, which the workflow does not
currently consume anywhere and which would be empty if the action still used the retired
`::set-output`.

## Verification

`git.version.sh` exercised in a container for every path:

| scenario | env | writes |
| --- | --- | --- |
| release | `SCRUTINY_VERSION=v0.9.3`, `CI=true`, ref master/branch | `v0.9.3` |
| tag push (docker images) | `CI=true`, `GITHUB_REF_NAME=v0.9.3`, `GITHUB_REF_TYPE=tag` | `v0.9.3` |
| branch push (nightly) | `CI=true`, `GITHUB_REF_NAME=master`, `GITHUB_REF_TYPE=branch` | `master#abcdef1` |
| empty override | `SCRUTINY_VERSION=`, tag push | `v0.9.4` (falls through, no bare `v`) |

And the workflow's extraction step:

| input | result |
| --- | --- |
| the real `version.go` | `ACCEPTED v0.9.2` |
| file with no `VERSION` const | `REJECTED` |
| missing file | `REJECTED` |
| `const VERSION = "1.10.0"` | `ACCEPTED v1.10.0` |

## AI disclosure

Per [AI_POLICY.md](AI_POLICY.md): this change was written by **Claude Code** (Opus 5). It diagnosed
the problem, confirmed it against the published v0.9.1 release asset and the published v0.9.1 docker
image, made the change, and ran the checks above in a container; nothing was installed on the host.

The one thing that cannot be exercised outside of GitHub Actions is the release workflow itself, so
the first real proof will be the next release.

That is the extent of the verification behind this PR as opened.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -61,7 +61,16 @@ jobs:
           version_metadata_path: ${{ github.event.inputs.version_metadata_path }}
       - name: Checkout release tag
         shell: bash
+        env:
+          # the packagr container's base image has no en_US.UTF-8 locale generated, and the
+          # runner injects LANG/LC_ALL=en_US.UTF-8 into every step regardless
+          LANG: C.UTF-8
+          LC_ALL: C.UTF-8
         run: |
+          # releasr and bumpr talk to the repo through go-git, which skips this check; this is the
+          # first step in the job to shell out to the real git CLI, so it's the first to hit it.
+          git config --global --add safe.directory "${GITHUB_WORKSPACE}"
+
           RELEASE_TAG="v${{ steps.bump_version.outputs.release_version }}"
           if ! git rev-parse -q --verify "refs/tags/${RELEASE_TAG}" > /dev/null; then
             # action-bumpr-go declares its output as `version` but emits `release_version`,
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -144,8 +144,14 @@ jobs:
         with:
           name: workspace
       - name: "Generate frontend version information"
+        env:
+          # the node:lts-slim image has no en_US.UTF-8 locale generated, and the
+          # runner injects LANG/LC_ALL=en_US.UTF-8 into every step regardless
+          LANG: C.UTF-8
+          LC_ALL: C.UTF-8
         run: |
           apt-get update && apt-get install -y git
+          git config --global --add safe.directory "${GITHUB_WORKSPACE}"
           cd webapp/frontend && chmod +x git.version.sh && ./git.version.sh
       - name: Build Frontend
         run: |
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -69,7 +69,7 @@ jobs:
             # releasr created if that ever stops arriving.
             RELEASE_TAG="$(git tag --points-at HEAD | head -n 1)"
           fi
-          if [[ ! "${RELEASE_TAG}" =~ ^v[0-9]+\.[0-9]+\.[0-9]+(-[0-9A-Za-z.-]+)?(\+[0-9A-Za-z.-]+)?$ ]]; then
+          if [[ ! "${RELEASE_TAG}" =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
             echo "could not determine the release tag (got '${RELEASE_TAG}')"
             exit 1
           fi
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -119,7 +119,15 @@ jobs:
         with:
           name: workspace
       - name: "Generate frontend version information"
-        run: "cd webapp/frontend && chmod +x git.version.sh && ./git.version.sh"
+        run: |
+          SCRUTINY_VERSION="v$(sed -n 's/^const VERSION = "\(.*\)"$/\1/p' "${{ github.event.inputs.version_metadata_path }}")"
+          if [[ ! "${SCRUTINY_VERSION}" =~ ^v[0-9]+\.[0-9]+\.[0-9]+ ]]; then
+            echo "could not read the release version from ${{ github.event.inputs.version_metadata_path }} (got '${SCRUTINY_VERSION}')"
+            exit 1
+          fi
+          export SCRUTINY_VERSION
+          cd webapp/frontend && chmod +x git.version.sh && ./git.version.sh
+        shell: bash
       - name: Build Frontend
         run: |
           apt-get update && apt-get install -y make
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -59,6 +59,22 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.SCRUTINY_GITHUB_TOKEN }} # Leave this line unchanged
         with:
           version_metadata_path: ${{ github.event.inputs.version_metadata_path }}
+      - name: Checkout release tag
+        shell: bash
+        run: |
+          RELEASE_TAG="v${{ steps.bump_version.outputs.release_version }}"
+          if ! git rev-parse -q --verify "refs/tags/${RELEASE_TAG}" > /dev/null; then
+            # action-bumpr-go declares its output as `version` but emits `release_version`,
+            # and it does so over the deprecated ::set-output. Fall back to the tag that
+            # releasr created if that ever stops arriving.
+            RELEASE_TAG="$(git tag --points-at HEAD | head -n 1)"
+          fi
+          if [[ ! "${RELEASE_TAG}" =~ ^v[0-9]+\.[0-9]+\.[0-9]+(-[0-9A-Za-z.-]+)?(\+[0-9A-Za-z.-]+)?$ ]]; then
+            echo "could not determine the release tag (got '${RELEASE_TAG}')"
+            exit 1
+          fi
+          echo "checking out ${RELEASE_TAG}"
+          git checkout --detach "refs/tags/${RELEASE_TAG}"
       - name: Upload workspace
         uses: actions/upload-artifact@v4
         with:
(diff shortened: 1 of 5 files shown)


kaysond:
Reworked per your review — the `SCRUTINY_VERSION` hand-off is gone and every workflow now derives the version the same way: **from the tag on the checked out commit**.

## `release.yaml`

New step after `Commit Changes Locally`, before the workspace upload:

```yaml
- name: Checkout release tag
  shell: bash
  run: |
    RELEASE_TAG="v${{ steps.bump_version.outputs.release_version }}"
    ...
    git checkout --detach "refs/tags/${RELEASE_TAG}"
```

Two things I found while wiring up the bumpr output that are worth knowing:

**The output is not called `version`.** `action-bumpr-go/action.yml` declares `outputs.version`, but the binary it downloads emits a *different* name — [`pipeline.go:93`](https://github.com/PackagrIO/bumpr/blob/v0.0.23/pkg/pipeline.go#L93) does `p.Scm.SetOutput("release_version", ...)`. Declared outputs are documentation only for Docker actions, so the real reference is `steps.bump_version.outputs.release_version`. `steps.bump_version.outputs.version` would have silently been the empty string.

**It is emitted over the deprecated `::set-output`.** [`scm_github.go:411`](https://github.com/PackagrIO/go-common/blob/master/scm/scm_github.go#L410-L413) is `fmt.Printf("\n::set-output name=%s::%s\n", ...)` — it has never moved to `$GITHUB_OUTPUT`. That still works today, but it is on GitHub's removal list and the release runs in the repo are old enough that their logs have expired, so I could not confirm it from a real run.

So rather than trust it blindly, the step falls back to the tag `releasr` created and validates either way:

```bash
if ! git rev-parse -q --verify "refs/tags/${RELEASE_TAG}" > /dev/null; then
  RELEASE_TAG="$(git tag --points-at HEAD | head -n 1)"
fi
if [[ ! "${RELEASE_TAG}" =~ ^v[0-9]+\.[0-9]+\.[0-9]+(-[0-9A-Za-z.-]+)?(\+[0-9A-Za-z.-]+)?$ ]]; then
  echo "could not determine the release tag (got '${RELEASE_TAG}')"
  exit 1
fi
```

`releasr` creates that tag locally at the release commit ([`engine_golang.go:57`](https://github.com/PackagrIO/releasr/blob/master/pkg/engine/engine_golang.go#L57), `git.GitTag(..., fmt.Sprintf("v%s", ...))`), and the workspace artifact already carries `.git` — `publishr` opens it with `git.PlainOpen`, so it has to.

**`node:lts-slim` has no git**, which the old env-var approach never needed. Verified:

```
$ docker run --rm node:lts-slim sh -c 'git --version || echo NO GIT'
sh: 1: git: not found
```

so the frontend job installs it before generating the version.

## `git.version.sh`

One rule, no `CI` branch any more:

```bash
GIT_TAG="$(git describe --tags --exact-match 2>/dev/null)"
if [[ "${GIT_TAG}" =~ ${SEMVER_TAG_REGEX} ]]; then
    VERSION_INFO="${GIT_TAG}"
else
    ...
    VERSION_INFO="${GIT_BRANCH}#$(git rev-parse --short=7 HEAD)"
fi
```

The branch fallback keeps the existing `master#d77976c` shape. `git rev-parse --abbrev-ref HEAD` returns `HEAD` when detached, so that case falls back to `GITHUB_REF_NAME`.

## Does `docker-build.yaml` still fire?

Yes — that was worth checking rather than assuming, because `actions/checkout` fetches tag pushes shallowly with `--no-tags`. It still creates the local tag ref through the explicit refspec, so `git describe --tags --exact-match` resolves. Reproduced exactly:

```
git fetch --no-tags --prune --no-recurse-submodules --depth=1 origin "+${SHA}:refs/tags/v0.9.4"
git checkout --progress --force refs/tags/v0.9.4
→ v0.9.4
```

## Verification

A harness drives the real `git.version.sh` and the real `Checkout release tag` logic against repos built to match each workflow's git state. Run in `node:lts-slim` + git, i.e. the release container:

```
== release.yaml ==
  PASS  bumpr output present -> tag checked out -> 'v0.9.3'
  PASS    local master ref still at release commit
  PASS    HEAD detached at the release commit
  PASS  bumpr output empty -> falls back to releasr's tag -> 'v0.9.3'
  PASS  no tag at all -> step fails
== docker-build.yaml (push of a semver tag) ==
  PASS  semver tag push -> version from tag -> 'v0.9.4'
== docker-nightly.yaml / branch builds ==
  PASS  untagged commit on master -> master#sha -> 'master#e2a2127'
  PASS  untagged commit on a branch -> branch#sha -> 'my_feature#cfabc2b'
== edge cases ==
  PASS  non-semver tag is ignored -> 'master#023ee4a'
  PASS  detached + untagged, no GITHUB_REF_NAME -> 'detached#fead903'
  PASS  detached + untagged, GITHUB_REF_NAME set -> 'some_branch#fead903'
  PASS  prerelease semver tag accepted -> 'v0.9.9-rc.1'
  PASS  lightweight tag accepted -> 'v0.9.10'
== artifact round trip (upload-artifact drops exec bits) ==
  PASS  version survives the workspace artifact -> 'v0.9.11'

PASS=14 FAIL=0
```

And against this repo rather than a synthetic one:

```
$ git checkout --detach v0.9.2 && ./git.version.sh
generating version file from the checked out release tag
writing version file (version: v0.9.2)

$ git checkout master && ./git.version.sh
no release tag on the checked out commit, generating version file from the branch
writing version file (version: master#a2f31ba)
```

## What I could not test

The release workflow itself. `workflow_dispatch` + `packagr` + a real tag push is not reproducible here, so the parts taken from reading source rather than running are: that `::set-output` is still honoured by the runner (hence the fallback), and that `git checkout --detach` does not disturb `publishr`. On that last point — `scmBase.Publish` builds its push branch with `GitCreateBranchFromHead` and pushes `refs/heads/<uuid>:refs/heads/master` plus the tag, so it uses HEAD's *commit*, which detaching does not move, and the local `master` ref is untouched. The harness asserts both of those, but only on a synthetic repo.

kaysond:
Rolled in two more fixes while testing this end-to-end on a real fork release run, both discovered by actually running the workflow:

## \`safe.directory\` / locale fix (release workflow)

Both git-shelling steps (\`Checkout release tag\`, and \`Generate frontend version information\` in the \`build_frontend\` job) failed with:

```
fatal: detected dubious ownership in repository at '/__w/scrutiny/scrutiny'
```

plus an unrelated \`setlocale: LC_ALL: cannot change locale (en_US.UTF-8)\` warning. Root cause: `packagrio/action-bumpr-go`/`action-releasr-go` talk to the repo through `go-git`, which skips git's ownership check entirely (CVE-2022-24765) — so neither of those earlier steps ever tripped it. `Checkout release tag` and the frontend job's version step are the first steps in the workflow to shell out to the real `git` CLI, so they're the first to hit it. Fixed with `git config --global --add safe.directory "${GITHUB_WORKSPACE}"` at the top of each step, plus a step-level `LANG`/`LC_ALL: C.UTF-8` override for the locale warning (the `packagrio/packagr` and `node:lts-slim` images don't have `en_US.UTF-8` generated, but the runner injects it into every step's env regardless).

## Docker image version banner shows \`dev-<version>\` for real releases

Unrelated to the workflow itself: none of the three Dockerfiles (`Dockerfile.web`, `Dockerfile`, `Dockerfile.collector`) ever set the `-X main.goos/main.goarch` ldflags the version banner depends on, so every docker image — including genuinely tagged releases — printed itself as `dev-0.9.3` instead of `linux.amd64-0.9.3`. Fixed by wiring BuildKit's standard `TARGETOS`/`TARGETARCH` build args into the existing `make` invocation in each Dockerfile's `backendbuild` stage. `Dockerfile` (omnibus) and `Dockerfile.collector` also needed `COLLECTOR_BINARY_NAME` pinned explicitly, since the Makefile appends `-$(GOOS)-$(GOARCH)` to whichever binary name isn't already overridden on the command line — without pinning it, this would've silently renamed the collector binary out from under the `COPY --from=backendbuild` steps.

Verified natively on linux/amd64 for both images (banner reads `linux.amd64-0.9.2` post-fix); could not verify an emulated arm64/arm build in this environment (QEMU cross-arch execution isn't available here), but the multi-arch build mechanism itself is unchanged — this only feeds two additional values into the already-working per-target-platform `RUN` step.

Was originally opened as a separate PR (#1044) before rolling it into this one, since both were found and fixed while shaking out the same release workflow run.

Written and tested with Claude Code (Opus 5), per [AI_POLICY.md](AI_POLICY.md).

kaysond:
Manually reviewed the changes and tested everything on my fork by doing a release and checking:
* collector binary (amd64)
* web binary (amd64)
* web frontend tar.gz (search for appVersion)
* docker web
* docker collector

PR #1044, opened 2026-08-07
Show the real build target in the docker image version banner

kaysond:
## Summary

The version banner printed by `scrutiny`/`scrutiny-collector-metrics` on startup falls back to a `dev-<version>` string whenever the `-X main.goos/main.goarch` ldflags aren't set at build time. None of the three Dockerfiles (`Dockerfile.web`, `Dockerfile`, `Dockerfile.collector`) ever set them, so every docker image - including genuinely tagged releases - reports itself as a dev build, e.g. `dev-0.9.3` instead of `linux.amd64-0.9.3`.

This adds `ARG TARGETOS`/`ARG TARGETARCH` (BuildKit's standard auto-populated platform args) to each `backendbuild` stage and passes them through to the existing `make` invocation. `Dockerfile` (omnibus) and `Dockerfile.collector` also needed `COLLECTOR_BINARY_NAME` pinned explicitly - the Makefile appends `-$(GOOS)-$(GOARCH)` to whichever binary name isn't already overridden on the command line, which would otherwise rename the collector binary out from under the `COPY --from=backendbuild` steps.

Multi-arch builds are unaffected by this change: the `backendbuild` stage was already running under QEMU per-target-platform before this change (that's how these images get built for arm64/arm today) - this only feeds two additional values into an already-working per-arch `RUN` step, it doesn't change what runs or how.

## Test plan

- [x] Built `Dockerfile.web` and `Dockerfile.collector` natively for `linux/amd64` via `docker buildx build`
- [x] Ran each resulting binary - banner now reads `linux.amd64-0.9.2` instead of `dev-0.9.2` for both web and collector images
- [ ] Emulated arm64/arm build (could not verify in this environment - QEMU cross-arch execution isn't available here)

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -61,7 +61,16 @@ jobs:
           version_metadata_path: ${{ github.event.inputs.version_metadata_path }}
       - name: Checkout release tag
         shell: bash
+        env:
+          # the packagr container's base image has no en_US.UTF-8 locale generated, and the
+          # runner injects LANG/LC_ALL=en_US.UTF-8 into every step regardless
+          LANG: C.UTF-8
+          LC_ALL: C.UTF-8
         run: |
+          # releasr and bumpr talk to the repo through go-git, which skips this check; this is the
+          # first step in the job to shell out to the real git CLI, so it's the first to hit it.
+          git config --global --add safe.directory "${GITHUB_WORKSPACE}"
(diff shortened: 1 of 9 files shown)


kaysond:
Rolled into #1040 — both fixes came out of the same fork release run, so they're bundled there instead of split across two PRs.

PR #803, opened 2025-09-15
Keep existing device type

thomashilzendegen:
This should fix #608 as it was overriding a configured device type from the `collector.yaml` with the (in my case wrong) results from the `smartctl --info` command (`device.type` JSON element).

PR #1067, opened 2026-08-23
fix(collector): honor a device type set in the config file

### LINKED ISSUE #1063: Scrutiny Collector Ignores Configured SCSI Device Type and Executes smartctl --device sat for USB Seagate BUP Portable Drives
### Discussed in https://github.com/AnalogJ/scrutiny/discussions/1057

<sup>Originally posted by **Lyfey** August 17, 2026</sup>
### Issue Description

Scrutiny successfully detects and monitors an internal NVMe drive but is unable to collect metrics from a USB-connected Seagate BUP Portable drive.

Direct smartctl testing indicates that the drive is accessible only when queried using the SCSI device type. However, Scrutiny appears to ignore the configured device type and continues executing SMART queries using SAT, causing collection to fail.

The issue appears to be within Scrutiny's device detection or command generation logic rather than Smartmontools itself.

### Expected Behavior

When a device is configured as:

YAML

```
devices:
  - device: /dev/sda
    type: scsi
```

Scrutiny should execute:
`smartctl --info --json --device scsi /dev/sda`

and use the returned device information to continue collecting metrics.

### Actual Behavior

Despite configuring the device as scsi in multiple ways, Scrutiny continues executing:

`smartctl --info --json --device sat /dev/sda`
or:
`smartctl --info --json /dev/sda`

which results in:

"exit_status": 2

Collector output:
```
INFO[0000] Verifying required tools type=metrics
INFO[0000] Executing command: smartctl --scan --json type=metrics
INFO[0000] Executing command: smartctl --info --json /dev/sda type=metrics
ERRO[0000] Could not retrieve device information for sda: exit status 2 type=metrics
```

or

```
INFO[0000] Verifying required tools type=metrics
INFO[0000] Executing command: smartctl --scan --json type=metrics
INFO[0000] Executing command: smartctl --info --json --device sat /dev/sda type=metrics
ERRO[0000] Could not retrieve device information for sda: exit status 2 type=metrics
```

### Reproduction Steps

1. Attach a Seagate BUP Portable USB drive.
2. Confirm the device appears as:
`   /dev/sda`
3. Verify Smartmontools behaviour:
 Fails:
`   smartctl --info --json /dev/sda`
 Fails:
`   smartctl --info --json --device sat /dev/sda`
 Succeeds:
`   smartctl --info --json --device scsi /dev/sda`
 
4. Configure Scrutiny in various ways - it is noted that some configurations listed below may be invalid
5. Restart Scrunity
6. Confirm Scrunity has configuration to be tested with:
`  docker exec -it scrutiny cat /opt/scrutiny/config/collector.yaml`
7. Confirm Scrunity results with
`  docker exec -it scrutiny scrutiny-collector-metrics run`
8. Scrutiny executes SAT or default detection instead of SCSI and collection fails.

--------- TESTING RESULTS -----------

`smartctl --info --json /dev/sda` and `smartctl --info --json --device sat /dev/sda`

```
{
  "device": {
    "name": "/dev/sda",
    "info_name": "/dev/sda [SAT]",
    "type": "sat",
    "protocol": "ATA"
  },
  "smartctl": {
    "exit_status": 2
  }
}
```

`smartctl --info --json --device scsi /dev/sda`

```
{
  "device": {
    "name": "/dev/sda",
    "type": "scsi",
    "protocol": "SCSI"
  },
  "scsi_vendor": "Seagate",
  "scsi_product": "BUP Portable",
  "model_name": "Seagate BUP Portable",
  "serial_number": "0000xxxHFR",
  "user_capacity": {
    "bytes": 4000787029504
  },
  "smart_support": {
    "available": true,
    "enabled": false
  },
  "smartctl": {
    "exit_status": 0
  }
}
```

Configuration Attempts:

---- CONFIG 1

```
version: 1
  devices:
  - device: /dev/sda
    type: scsi
```

Note - a sub-config of type: 'scsi' also applies
Results in - 
  Executing command: smartctl --info --json **{placeholder for attention}** /dev/sda  type=metrics

---- CONFIG 2

```
version: 1
commands:
  metrics_smartctl_args: "--device=scsi"
devices:
  - device: /dev/nvme0
    type: nvme
  - device: /dev/sda
    type: scsi
```

Results in - 
Executing command: smartctl --info --json **{placeholder for attention}** /dev/sda  type=metrics

---- CONFIG 3

```
version: 1
smartctl:
  devices:
    - device: /dev/sda
      type: scsi
      ignore: true
    - device: /dev/nvme0
      type: nvme
```

Results in - 
Executing command: smartctl --info --json **--device sat** /dev/sda  type=metrics

---- CONFIG 4

```
version: 1
smartctl:
  devices:
    - device: /dev/sda
      type: scsi
    - device: /dev/nvme0
      type: nvme
```

Results in -   
Executing command: smartctl --info --json **--device sat** /dev/sda  type=metrics
 
 ---- CONFIG 5

```
version: 1
commands:
  metrics_info_args: '-d scsi'
  metrics_smart_args: '-d scsi'
smartctl:
  devices:
    - device: /dev/sda
      type: scsi
    - device: /dev/nvme0
      type: nvme
```

Results in -   
Executing command: smartctl --info --json **--device sat** /dev/sda  type=metrics
 
 ### scrutiny debug logs

```text
2026/08/17 07:15:05 Loading configuration file: /opt/scrutiny/config/collector.yaml

 ___   ___  ____  __  __  ____  ____  _  _  _  _
/ __) / __)(  _ \(  )(  )(_  _)(_  _)( \( )( \/ )
\__ \( (__  )   / )(__)(   )(   _)(_  )  (  \  /
(___/ \___)(_)\_)(______) (__) (____)(_)\_) (__)
AnalogJ/scrutiny/metrics                                dev-0.8.2

INFO[0000] Verifying required tools                      type=metrics
INFO[0000] Executing command: smartctl --scan --json     type=metrics
INFO[0000] Executing command: smartctl --info --json /dev/sda  type=metrics
ERRO[0000] Could not retrieve device information for sda: exit status 2  type=metrics
INFO[0000] Executing command: smartctl --info --json --device nvme /dev/nvme0  type=metrics
INFO[0000] Using WWN Fallback                            type=metrics
INFO[0000] Sending detected devices to API, for filtering & validation  type=metrics
INFO[0000] Collecting smartctl results for nvme0         type=metrics
INFO[0000] Executing command: smartctl --xall --json --device nvme /dev/nvme0  type=metrics
INFO[0000] Publishing smartctl results for 212646803411  type=metrics
INFO[0000] Main: Completed                               type=metrics
```

### Scrutiny Version

dev-0.8.2

### Smartmontools Version

smartctl 7.5 2 svn revision 5714

### OS Version Information

PRETTY_NAME="Ubuntu 26.04 LTS"

### Component

collector

### Deployment Method

- [X] docker
- [ ] binaries
- [ ] systemd

### Hard Drive Information

Model: Seagate BUP Portable
Capacity: 4,000,787,029,504 bytes (4 TB)
Connection: USB
Device: /dev/sda

### smartctl output

Successful command - `smartctl --info --json --device scsi /dev/sda`

```json
{
  "json_format_version": [
    1,
    0
  ],
  "smartctl": {
    "version": [
      7,
      5
    ],
    "pre_release": false,
    "svn_revision": "5714",
    "platform_info": "x86_64-linux-7.0.0-29-generic",
    "build_info": "(local build)",
    "argv": [
      "smartctl",
      "--info",
      "--json",
      "--device=scsi",
      "/dev/sda"
    ],
    "exit_status": 0
  },
  "local_time": {
    "time_t": 1786950731,
    "asctime": "Mon Aug 17 07:12:11 2026 UTC"
  },
  "device": {
    "name": "/dev/sda",
    "info_name": "/dev/sda",
    "type": "scsi",
    "protocol": "SCSI"
  },
  "scsi_vendor": "Seagate",
  "scsi_product": "BUP Portable",
  "model_name": "Seagate BUP Portable",
  "scsi_model_name": "Seagate BUP Portable",
  "scsi_revision": "0004",
  "scsi_version": "SPC-4",
  "user_capacity": {
    "blocks": 7814037167,
    "bytes": 4000787029504
  },
  "logical_block_size": 512,
  "physical_block_size": 4096,
  "logical_unit_id": "0x5000000000000001",
  "serial_number": "0000xxxHFR",
  "device_type": {
    "scsi_terminology": "Peripheral Device Type [PDT]",
    "scsi_value": 0,
    "name": "disk"
  },
  "smart_support": {
    "available": true,
    "enabled": false
  },
  "temperature_warning": {
    "enabled": false
  }
```

Failing commands `smartctl --info --json /dev/sda` and `smartctl --info --json --device sat /dev/sda`

```
{
  "json_format_version": [
    1,
    0
  ],
  "smartctl": {
    "version": [
      7,
      5
    ],
    "pre_release": false,
    "svn_revision": "5714",
    "platform_info": "x86_64-linux-7.0.0-29-generic",
    "build_info": "(local build)",
    "argv": [
      "smartctl",
      "--info",
      "--json",
      "--device",
      "sat",
      "/dev/sda"
    ],
    "exit_status": 2
  },
  "local_time": {
    "time_t": 1786952684,
    "asctime": "Mon Aug 17 07:44:44 2026 UTC"
  },
  "device": {
    "name": "/dev/sda",
    "info_name": "/dev/sda [SAT]",
    "type": "sat",
    "protocol": "ATA"
  }
}
```

### docker-compose.yml

```yaml
services:
  scrutiny:
    container_name: scrutiny
    image: ghcr.io/analogj/scrutiny:master-omnibus
    restart: unless-stopped

    ports:
      - "8081:8080"

    privileged: true

    cap_add:
      - SYS_RAWIO
      - SYS_ADMIN

    volumes:
      - /run/udev:/run/udev:ro
      - /srv/docker/scrutiny:/opt/scrutiny/config

    devices:
      - /dev/sda:/dev/sda
      - /dev/nvme0:/dev/nvme0
```

### scrutiny.yaml

```yaml
empty
```

### collector.yaml

```yaml
Tested configurations listed under Reproduction steps. Expected working configuration:

Version: 1
  devices:
  - device: /dev/sda
    type: scsi
```

_No response_

### I acknowledge that:

justadityaraj:
Fixes #1063.

## What was wrong

The scsi/ata suppression keyed off the device type's **value**, so it could not tell a type that came from `smartctl --scan` from one the user pinned in `collector.yaml`:

```go
if len(device.DeviceType) > 0 && device.DeviceType != "scsi" && device.DeviceType != "ata" {
    args = append(args, "--device", device.DeviceType)
}
```

The suppression itself is correct and stays — the comment explains it, ata drives get scanned as scsi under docker and metadata is lost. The problem is that `TransformDetectedDevices` writes a config override into the *same* `models.Device.DeviceType` field the scanner writes to, so provenance is already gone by the time the guard runs.

So `devices: [{device: /dev/sda, type: scsi}]` produced a bare `smartctl --info --json /dev/sda`, smartctl auto-selected SAT, exited 2, and the drive was dropped with no device info. `ValidateConfig` also rejects `-d` inside `metrics_info_args`, so there is no user-side workaround.

**Both call sites need it.** `detect.go` builds `--info`, `metrics.go` builds `--xall`. Fixing only `detect.go` gets the device registered and then `Collect()` drops the type again — I confirmed that second drop with its own failing test.

## The fix

`HasDeviceTypeOverride` asks the config whether the type was pinned, so suppression now applies only to a scanned type. It is a method on `config.Interface`, sharing one `getDeviceOverride` lookup with `GetCommandMetricsInfoArgs` and `GetCommandMetricsSmartArgs`, and the checked-in gomock is updated to match.

The guard reads `(type != "scsi" && type != "ata") || HasDeviceTypeOverride(...)`, so the config is only consulted for a type that is already scsi/ata. Every other device keeps exactly the old code path, and the existing strict-gomock tests make no unexpected call.

## Verification

**End to end, with the real collector binary** and a stub `smartctl` on `PATH` that logs its argv, against `devices: [{device: /dev/sda, type: scsi}]`:

```
before:  Executing command: smartctl --info --json /dev/sda
after:   Executing command: smartctl --info --json --device scsi /dev/sda
```

The "before" line is byte-for-byte the log in the issue, and the "after" line is exactly the expected behaviour it asks for.

**Tests** — table tests at both call sites. Against the unpatched tree they fail with:

```
detect:   Got: [--info --json /dev/sda]   Want: [--info --json --device scsi /dev/sda]
metrics:  Got: [--xall --json /dev/sda]   Want: [--xall --json --device scsi /dev/sda]
```

They cover the cases that must **not** change too: a scanned `sat` still passes through, a scanned `scsi` is still dropped, an override on a different device doesn't leak, and an override with no `type` set doesn't count. `go test ./collector/...` is green, `go vet` and `gofmt -l` clean.

## Re-verified after the rebase

Against the rebased head, driving the real `config` package and the real `Detect` with a capturing shell and the issue's exact config (`devices: [{device: /dev/sda, type: scsi}]`) and scan result (`/dev/sda`, scanned type `sat`):

```
master 5bf0b25:  smartctl --scan --json
                 smartctl --info --json /dev/sda
this branch:     smartctl --scan --json
                 smartctl --info --json --device scsi /dev/sda
```

The configured `scsi` beats the scanned `sat`, which is the behaviour #803 now preserves through `SmartCtlInfo`.

Behavioural equivalence of the shared lookup was checked the same way: `GetCommandMetricsInfoArgs` and `GetCommandMetricsSmartArgs` return byte-identical results to master across duplicate `device:` entries and mixed-case paths.

`go test ./collector/...` fails the same five pre-existing path tests as master on a Windows host, no more and no fewer; `go vet` and `gofmt -l` are clean.

## AI usage disclosure

Per [AI_POLICY.md](https://github.com/AnalogJ/scrutiny/blob/master/AI_POLICY.md): this was written with **Claude Code**, used substantially — for tracing the two call sites, drafting the patch and the tests, and writing this description.

On the "fully verified with human use" requirement, precisely what I did and did not do: the fix is verified by running the actual `collector-metrics` binary and reading the command it executed, before and after, plus the tests above — not by reasoning about the code. I do **not** have the reporter's USB Seagate BUP drive, so the final `smartctl` exit-status-2 behaviour on that specific hardware is not something I can reproduce; what I verified is the argv scrutiny constructs, which is the part this changes. Happy to adjust if you'd rather see it done differently.

### DIFF
--- a/collector/pkg/collector/metrics.go
+++ b/collector/pkg/collector/metrics.go
@@ -137,10 +137,10 @@ func (mc *MetricsCollector) Collect(scrutiny_uuid uuid.UUID, deviceName string, 
 
 	fullDeviceName := fmt.Sprintf("%s%s", detect.DevicePrefix(), deviceName)
 	args := strings.Split(mc.config.GetCommandMetricsSmartArgs(fullDeviceName), " ")
-	//only include the device type if its a non-standard one, or the user pinned it in the config file.
+	//only include the device type if its a non-standard one, or the user set it in the config file.
 	//In some cases ata drives are detected as scsi in docker, and metadata is lost, so a scanned scsi/ata type is dropped.
 	if len(deviceType) > 0 &&
-		((deviceType != "scsi" && deviceType != "ata") || config.HasDeviceTypeOverride(mc.config, fullDeviceName)) {
+		((deviceType != "scsi" && deviceType != "ata") || mc.config.HasDeviceTypeOverride(fullDeviceName)) {
 		args = append(args, "--device", deviceType)
 	}
 	args = append(args, fullDeviceName)
--- a/collector/pkg/collector/metrics.go
+++ b/collector/pkg/collector/metrics.go
@@ -137,8 +137,10 @@ func (mc *MetricsCollector) Collect(scrutiny_uuid uuid.UUID, deviceName string, 
 
 	fullDeviceName := fmt.Sprintf("%s%s", detect.DevicePrefix(), deviceName)
 	args := strings.Split(mc.config.GetCommandMetricsSmartArgs(fullDeviceName), " ")
-	//only include the device type if its a non-standard one. In some cases ata drives are detected as scsi in docker, and metadata is lost.
-	if len(deviceType) > 0 && deviceType != "scsi" && deviceType != "ata" {
+	//only include the device type if its a non-standard one, or the user pinned it in the config file.
+	//In some cases ata drives are detected as scsi in docker, and metadata is lost, so a scanned scsi/ata type is dropped.
+	if len(deviceType) > 0 &&
+		((deviceType != "scsi" && deviceType != "ata") || config.HasDeviceTypeOverride(mc.config, fullDeviceName)) {
 		args = append(args, "--device", deviceType)
 	}
 	args = append(args, fullDeviceName)
--- a/collector/pkg/collector/metrics_test.go
+++ b/collector/pkg/collector/metrics_test.go
@@ -1,9 +1,18 @@
 package collector
 
 import (
-	"github.com/stretchr/testify/require"
+	"errors"
 	"net/url"
 	"testing"
+
+	mock_shell "github.com/analogj/scrutiny/collector/pkg/common/shell/mock"
+	mock_config "github.com/analogj/scrutiny/collector/pkg/config/mock"
+	"github.com/analogj/scrutiny/collector/pkg/detect"
+	"github.com/analogj/scrutiny/collector/pkg/models"
+	"github.com/gofrs/uuid/v5"
+	"github.com/sirupsen/logrus"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/mock/gomock"
 )
 
 func TestApiEndpointParse(t *testing.T) {
@@ -35,4 +44,70 @@ func TestApiEndpointParse_WithBasepathWithTrailingSlash(t *testing.T) {
 
 	url2, _ := baseURL.Parse("/d/e")
 	require.Equal(t, "http://localhost:8080/d/e", url2.String())
+}
+
+func TestMetricsCollector_Collect_DeviceType(t *testing.T) {
+	for _, tt := range []struct {
+		name         string
+		deviceType   string
+		overrides    []models.ScanOverride
+		expectedArgs []string
+	}{
+		{
+			name:         "should drop a scanned scsi type",
+			deviceType:   "scsi",
+			overrides:    []models.ScanOverride{},
+			expectedArgs: []string{"--xall", "--json"},
+		},
+		{
+			name:       "should keep a configured scsi type",
+			deviceType: "scsi",
+			overrides: []models.ScanOverride{
+				{Device: detect.DevicePrefix() + "sda", DeviceType: []string{"scsi"}},
+			},
+			expectedArgs: []string{"--xall", "--json", "--device", "scsi"},
+		},
+	} {
+		t.Run(tt.name, func(t *testing.T) {
+			ctrl := gomock.NewController(t)
+
+			const (
+				someArgs       = "--xall --json"
+				someDeviceName = "sda"
+			)
+
+			fullDeviceName := detect.DevicePrefix() + someDeviceName
+
+			fakeConfig := mock_config.NewMockInterface(ctrl)
+			fakeConfig.EXPECT().
+				GetCommandMetricsSmartArgs(fullDeviceName).
+				Return(someArgs)
+			fakeConfig.EXPECT().
+				GetString("commands.metrics_smartctl_bin").
+				Return("smartctl")
+			fakeConfig.EXPECT().
+				GetDeviceOverrides().
+				Return(tt.overrides).
+				AnyTimes()
+
+			someLogger := logrus.WithFields(logrus.Fields{})
+
+			fakeShell := mock_shell.NewMockInterface(ctrl)
+			fakeShell.EXPECT().
+				Command(someLogger, "smartctl", append(tt.expectedArgs, fullDeviceName), "", gomock.Any()).
+				Return("", errors.New("smartctl unavailable in test"))
+
+			apiEndpoint, err := url.Parse("http://localhost:8080/")
+			require.NoError(t, err)
+
+			mc := MetricsCollector{
+				config:        fakeConfig,
+				BaseCollector: BaseCollector{logger: someLogger},
+				apiEndpoint:   apiEndpoint,
+				shell:         fakeShell,
+			}
+
+			mc.Collect(uuid.Must(uuid.NewV4()), someDeviceName, tt.deviceType)
+		})
+	}
 }
--- a/collector/pkg/collector/metrics_test.go
+++ b/collector/pkg/collector/metrics_test.go
@@ -8,7 +8,6 @@ import (
 	mock_shell "github.com/analogj/scrutiny/collector/pkg/common/shell/mock"
 	mock_config "github.com/analogj/scrutiny/collector/pkg/config/mock"
 	"github.com/analogj/scrutiny/collector/pkg/detect"
-	"github.com/analogj/scrutiny/collector/pkg/models"
 	"github.com/gofrs/uuid/v5"
 	"github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/require"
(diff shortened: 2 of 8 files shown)


kaysond (changes requested):
Thanks for putting this together. Two requests:

1. the method added to config.go should be an actual method, not a pkg func (e.g. `func (c *Config) Has...`), and we should not use strings.EqualFold, we should do an O(1) lookup on the device overrides keys, because it should be a map. (this might need some extra work, which should live in its own commit)
2. I like adding tests, but it's way too many lines and too verbose with the struct, as LLMs like to do

justadityaraj:
Sure, doing the 1.
2- yes, took some help from Codex.

justadityaraj:
Both done, in two commits.

`6d34caa` is the lookup change on its own, as you suggested: the overrides are parsed once into a map keyed on the lowercased device name, `HasDeviceTypeOverride` is a method on the config reading that map instead of a package function, and `GetDeviceOverrides` still returns the slice so ordering is unchanged. Matching stays case-insensitive and a device listed twice still resolves to its first entry. The mock is regenerated for the new interface method.

`c405ebd` cuts the tests down — 108 lines removed against 26 added across `detect_test.go` and `metrics_test.go`. What's left is the cases that actually distinguish an unset device type from an explicitly configured one.

`go vet ./collector/...` and `gofmt` are clean; `collector/pkg/config` and `collector/pkg/collector` pass. `collector/pkg/detect` has four failures on this Windows host, but they reproduce identically on `master` — they assert `/dev/sda` against a path the OS renders differently.

kaysond:
@justadityaraj - I rebased this on some changes I merged. decided to keep things as case-insensitive, for now, since its not consistent elsewhere. can you please do a quick sanity check?

kaysond (approved):

justadityaraj:
Checked it, rebase looks right. Both of your edits hold up, and one of them was load-bearing in a way I had not accounted for.

**The fix still works end to end.** Driving the real `config` package and the real `Detect` with a capturing shell, using the issue's exact config (`devices: [{device: /dev/sda, type: scsi}]`) and its scan result (`/dev/sda`, scanned type `sat`):

```
master 5bf0b25:  smartctl --info --json /dev/sda
6dbe074:         smartctl --info --json --device scsi /dev/sda
```

The first line is byte-for-byte the log in #1063. The configured `scsi` correctly beats the scanned `sat`.

**The operand reorder was not cosmetic, it was needed.** `HasDeviceTypeOverride` is a pure read, so `(type != "scsi" && type != "ata") || Has(x)` is result-identical to what I had. But your ordering also keeps the config off the path for every non-scsi/ata device, which is why `.AnyTimes()` is now required rather than papering over anything: in the `megaraid,14` case the call genuinely never happens. I checked the tests are still load-bearing by reverting only the guard at both call sites to master's form:

```
--- FAIL: TestDetect_SmartCtlInfo_ConfiguredDeviceType/should_keep_a_configured_scsi_type
--- FAIL: TestMetricsCollector_Collect_ConfiguredDeviceType/should_keep_a_configured_scsi_type
        Want: [--xall --json --device scsi sda]
        Got:  [--xall --json sda]
```

The other two cases pass under that mutation, which is correct: they exist to pin the behaviour that must *not* change.

**`EqualFold` over the map reads fine to me**, and it is the more consistent choice next to #1073. I confirmed the extraction into `getDeviceOverride` is behaviour-preserving rather than assuming it: `GetCommandMetricsInfoArgs` and `GetCommandMetricsSmartArgs` return byte-identical results to master across duplicate `device:` entries and mixed-case paths, including that first-match-wins is unchanged.

**#803 composes cleanly, and makes the `metrics.go` half necessary.** Now that `SmartCtlInfo` no longer overwrites `device.DeviceType` with smartctl's resolved transport, a pinned `scsi` survives detection and arrives at `Collect()` still as `scsi`, so it needs the same guard there. Before #803 it would often have been rewritten to `sat` and passed through on the non-standard branch by accident.

`go test ./collector/...` at the head fails the same five pre-existing path tests as master on my Windows host, no more and no fewer. `go vet` and `gofmt -l` clean, all 15 checks green.

One thing worth a release note, not a code change: anyone who already has `type: scsi` or `type: ata` in their `collector.yaml` will start getting a `--device` flag they have never been sent. That is the point of the fix, but the suppression you are relaxing was added precisely because `-d scsi` can lose metadata on ata drives that docker detects as scsi, so it seems worth a line rather than a silent change. None of the `type:` examples in `example.collector.yaml` are scsi/ata, so the blast radius is only configs that pin one deliberately.

I also fixed the PR description: it still claimed there was no `config.Interface` change and no mock regeneration, which stopped being true in `6d34caa`, and the short-circuit paragraph described my ordering rather than yours.

Same disclosure as before, per AI_POLICY.md: this check was done with Claude Code. The before/after commands above are real runs against the two trees, not reasoning about the code. I still do not have the reporter's USB Seagate drive, so smartctl's exit status on that specific hardware remains the one thing I cannot confirm.

PR #1090, opened 2026-09-14
feat(collector): add built-in cron scheduling with run --cron

### LINKED ISSUE #956: Add built-in cron support to the collector
Facilitates creation of distroless base images.

This should probably check for the same env vars (`COLLECTOR_CRON_SCHEDULE`, `COLLECTOR_RUN_STARTUP`, `COLLECTOR_RUN_STARTUP_SLEEP`) for backwards compatibility in addition to supporting cli args and yaml conf.

anandghegde:
Closes #956

Adds a built-in scheduler to the collector, so it can run on a schedule without system cron (e.g. in a distroless image).

```
scrutiny-collector-metrics run --cron --cron-schedule "0 */6 * * *" --run-startup
```

- `--cron` keeps the collector running and calls the normal collection on each tick.
- `--cron-schedule`, `--run-startup`, `--run-startup-sleep` read `COLLECTOR_CRON_SCHEDULE`, `COLLECTOR_RUN_STARTUP` and `COLLECTOR_RUN_STARTUP_SLEEP`, the same variables the docker images use, with the same defaults (`0 0 * * *`, false, 1). They can also go in a new `cron:` section of `collector.yaml`.
- Runs never overlap: the next tick is worked out after a run finishes. A failed run is logged and the schedule keeps going, same as cron.
- Surrounding quotes on the schedule are stripped, like the entrypoint scripts do.
- Parsing uses `github.com/robfig/cron/v3` (MIT, no dependencies of its own). It takes standard 5-field expressions plus `@daily` / `@every 6h`. Only the parser is used, not its job runner.

**Why `--cron` is opt-in instead of switching on when `COLLECTOR_CRON_SCHEDULE` is set:** both current images pass the container env to the collector runs they start. The collector image sources `/env.sh` in the crontab line, and omnibus uses `with-contenv` in `collector-once`. If the env var alone started the scheduler, every cron-started `run` for anyone who sets `COLLECTOR_CRON_SCHEDULE` would never exit. With `--cron` required, plain `run` behaves exactly as before, and a future image can use `CMD ["scrutiny-collector-metrics", "run", "--cron"]`. The existing Dockerfiles/s6 scripts aren't touched.

Docs: updated the "Cron Schedule" section of the README (it said the schedule couldn't be set in `collector.yaml`) and added the `cron:` block to `example.collector.yaml`.

### Testing

Unit tests:
- `ParseCronSchedule`: standard expressions, quoted values, descriptors, invalid input
- `RunOnSchedule`: repeated runs with a short fake schedule that keep going after errors, the startup run on and off, and stopping
- config: defaults, values from `collector.yaml`, and `COLLECTOR_CRON_SCHEDULE` overriding the file

Commands run on macOS (arm64):
- `go test ./...` with the `influxdb:2.8` container from CONTRIBUTING: all packages pass
- `go vet ./collector/...`, gofmt clean. `golangci-lint run ./collector/...` shows the same 40 issues as on `master` and none new.

Manual runs of the built collector against a local `latest-omnibus` container, with smartmontools 7.5 from Homebrew:
- `COLLECTOR_RUN_STARTUP=true COLLECTOR_RUN_STARTUP_SLEEP=2 run --cron --cron-schedule "* * * * *"`: startup run after 2s, then a full collection at the top of each minute (16:36:00, 16:37:00, 16:38:00) that published results to the API. Exited on SIGTERM.
- `COLLECTOR_CRON_SCHEDULE="'@every 10s'" run --cron --config cron.yaml` where the yaml sets `schedule: '0 0 1 1 *'`, `run_startup: true`, `run_startup_sleep: 3`. Result: startup run 3s after start, the quoted env schedule overrode the file, and a run taking about 40s was followed by the next one 10s after it finished (no overlap).
- `run --cron --cron-schedule "61 * * * *"`: exits 1 with `invalid cron schedule "61 * * * *": end of range (61) above maximum (59)`.
- `COLLECTOR_CRON_SCHEDULE="@every 10s" run` (no `--cron`): one collection, then exit 0, as before.
- linux/arm64 static build as PID 1 in `gcr.io/distroless/static-debian12` with `run --cron --cron-schedule "@every 5s" --run-startup` (no smartctl, so each run logs `smartctl binary is missing` and the schedule continues). `docker stop` returned in under a second.

Not tested: a real Linux host with SATA/SAS disks, or changes to the existing docker images (there aren't any).

### AI disclosure

This PR was written with Claude Code: the implementation, tests, docs, and the verification runs above.
I also ran it myself on my Mac (smartmontools from Homebrew, `latest-omnibus` container on :8080):
- `run --cron --cron-schedule "* * * * *" --run-startup`: the startup collection started right away (device scan and `smartctl --info` per disk).
- `run --cron --cron-schedule "@every 30s"`: logged `Next collector run scheduled for …:09:33`, ran a full collection at that time and published results for 4 of 5 detected devices (`Main: Completed`), then logged the next run 30s after that collection finished (`…:10:15`). Ctrl+C stopped it.

The per-device smartctl errors in those runs (internal Apple NVMe path not queryable, checksum warnings) come from smartctl on macOS in the existing device detection/collection code, which this PR doesn't touch.


kaysond:
Thanks for submitting this. Promise to get to it eventually but it might take a while

anandghegde:
No rush at all — thanks for the heads-up. The branch is rebased on current master and the collector tests pass, so it should still apply cleanly whenever you get to it. Ping me if you'd like anything split out or reshaped in the meantime.

